PR #28, opened 2024-10-02
Add containing_rectangle graph connection method for m2g

joeloskarsson:
## Describe your changes
This PR introduces a new method for connecting a mesh graph to the grid.

### Background
In earlier work we have used 4-nearest-neighbors (4nn) to create the m2g connection. This is not terrible, but has its issues (see  below), especially for grid cells along the edges (outside the area covered by the mesh graph).

Below is an example of how a 4nn m2g graph can look like, with a few edges highlighted

4nn_problem_illustration

* In red: These edges are not really desirable, as they represent strangely long connections in the graph
* In green: This edge does not exist, but it would be natural if it would. This mesh node represents one corner of the "rectangle" made up of 4 mesh nodes.

### containing_rectangle
The 4nn method was somewhat inspired by how m2g is created in GraphCast, where each grid node is projected up to the corresponding triangular face containing it in the mesh graph. The  `containing_rectangle` method is an attempt at achieving something closer to this for regular quadrilateral mesh graphs.

Starting with the end result, m2g edges created using the `containing_rectangle` look like

cont_example

The idea of this method is to similarly to GraphCast find the face of the mesh graph containing each grid node and connect the grid nodes with the edges of that face. Since we currently work with a regular layout for the mesh nodes these faces are rectangles, as illustrated below:

cont_example_grid

Note that this means that grid nodes along the edges will have less than 4 mesh connections.

### Definition of edge set

The actual definition of this edge set is as follows: Let $d_x$ be the distance between the columns of mesh nodes in the x-direction and similarly $d_y$ in the y-direction. A mesh nodes with coordinates $(x_m, y_m)$ should be connected to a grid node with coordinates $(x_g, y_g)$ if 

$$
(|x_g - x_m| < d_x) ~\\&~ (|y_g - y_m| < d_y).
$$

To avoid having to check all grid and mesh nodes pairwise for the expression above we  in practice start by building a radius graph based on radius $r = \sqrt{d_x^2 + d_y^2}$. This gives an edge set that is a superset of the `containing_rectangle` edge set. We then filter this edge set using the expression above to get the final set of edges. Trying to build a KD-tree using this distance measure directly seemed a bit over-complicated and this two-step method works fast without any issues. 

For now this method is only defined for m2g. It is not entirely clear how to do it for g2m, but it should be possible to extend also to those edges.

## Issue Link

No relevant issue.

## Type of change

  - As far as I see there is no place in the docs where the different graph connections are discussed, so will not add that right now.

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [ ] the tests pass

### DIFF
--- a/src/weather_model_graphs/create/base.py
+++ b/src/weather_model_graphs/create/base.py
@@ -234,6 +234,9 @@ def connect_nodes_across_graphs(
             mesh_node_dx = mesh_node_dx[0]
             mesh_node_dy = mesh_node_dy[0]
 
+        # This function is a filter that applies to edges, represented as vectors (vx, vy) in R^ 2.
+        # The filter is True if |vx| < dx & |vy| < dy, where dx and dy are the distance between
+        # rows and columns in source quadrilateral graph.
         def _edge_filter(edge_prop):
             abs_diffs = np.abs(edge_prop["vdiff"])
             return abs_diffs[0] < mesh_node_dx and abs_diffs[1] < mesh_node_dy
--- a/src/weather_model_graphs/create/base.py
+++ b/src/weather_model_graphs/create/base.py
@@ -212,7 +212,8 @@ def connect_nodes_across_graphs(
         # Connect to all nodes that could potentially be close enough,
         # which is at a relative distance of 1
         rad_graph = connect_nodes_across_graphs(
-            G_source, G_target, method="within_radius", rel_max_dist=1.)
+            G_source, G_target, method="within_radius", rel_max_dist=1.0
+        )
 
         # Filter edges to those that fit within a rectangle of measurements dx,dy
         mesh_node_dx = G_source.graph["dx"]
@@ -228,8 +229,11 @@ def connect_nodes_across_graphs(
             abs_diffs = np.abs(edge_prop["vdiff"])
             return abs_diffs[0] < mesh_node_dx and abs_diffs[1] < mesh_node_dy
 
-        filtered_edges = [(u, v) for u, v, edge_prop in rad_graph.edges(data=True)
-                if _edge_filter(edge_prop)]
+        filtered_edges = [
+            (u, v)
+            for u, v, edge_prop in rad_graph.edges(data=True)
+            if _edge_filter(edge_prop)
+        ]
 
         filtered_graph = rad_graph.edge_subgraph(filtered_edges)
         return filtered_graph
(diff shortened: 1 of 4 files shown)


TomasLandelius:
Just a thought - what about requiring that the mesh surrounds the grid so that all grid points have 4 decent nns?

src/weather_model_graphs/create/base.py
```diff
@@ -198,7 +205,52 @@ def connect_nodes_across_graphs(
...
+                "to use `containing_rectangle` you should not set `max_dist`, `rel_max_dist`or `max_num_neighbours`"
+            )
+        assert (
+            "dx" in G_source.graph and "dy" in G_source.graph
+        ), "Source graph must have dx and dy properties to connect nodes using method containing_rectangle"
+
+        # Connect to all nodes that could potentially be close enough,
+        # which is at a relative distance of 1
+        rad_graph = connect_nodes_across_graphs(
+            G_source, G_target, method="within_radius", rel_max_dist=1.0
```
sadamov, line 224: Why is this 1.0 and not `srt(dx^2 + dy^2)`?

sadamov (approved):
Thanks a lot for this extension. The code is in good shape. I have some conceptual questions and one technical comment (mostly for my education I guess).

- Once we start connecting realistic boundaries the state: is your statement below referring to the nodes at the border of the state or at the border of the boundary(+state)? I don't think we would want irregularities neither in grid nor mesh nodes at the border of the state, right?
`The outermost grid nodes having a somewhat strange and irregular connectivity pattern (current situation), or
The outermost mesh nodes having a somewhat strange and irregular connectivity pattern`
- Could `containing_rectangle` be used for many shapes in the future (e.g. hex-grid). Will we add seperate functions for each type of grid or call this one `containing_shape`?

- Looking at the roadmap I saw that graphtools is planned as an additional backend for v0.3.0. The motivation for the two-step approach (within-radius -> containing-rectangle) seems to be performance. I was wondering whether you would still do this in a highly optimized library like graphtools?

src/weather_model_graphs/create/base.py
joeloskarsson: Good question. This is relative distance, which is relative to the longest edge in (the bottom level of) the source and target graphs. These longest edges will be the diagonal edges in each rectangle in the mesh graph, which have the length ` sqrt(dx^2 + dy^2)`. So that is exactly the distance of rel_max_dist=1.0. 

I added a bit of a clarification about this, and also realized that there was nothing describing the `rel_max_dist` argument in the docstring, so fixed that as well.

joeloskarsson:
@TomasLandelius did you want to go over this code in more detail or are you happy with the addition?

TomasLandelius:
@joeloskarsson Sorry for being so slow. I'm happy with this!

TomasLandelius (approved):

PR #31, opened 2024-10-09
Make save function more universal to accept any number of 1D or 2D node or edge features

maxiimilian:
## Describe your changes

This pull request contains three changes -- one major and two minor changes.

The major change modifies how edge and node features are converted from `pyg.data.Data` objects into `torch.Tensor`s before saving. If we assume that all node/edge features (features in the following) are numeric, a conversion to float32 will always work. Further, my proposed change assumes that the features can be either 1D (vector of edge lengths) or 2D (node position, vdiff vector between source and target nodes). Based on this, I propose a universal function that reshapes any 1D vector into a 2D column vector, which allows the concatenation of any number of features in a loop. 

Using the new function, we also make the saving of `len` as edge feature explicit by including it in the default `edge_features` list. 

Two minor bugs are fixed:
- Default arguments of functions should not be mutables (list, dicts, any object). See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.
- The `hide_ticks` argument for `nx.draw` seems to be deprecated because it caused an error with my networkx v 3.2

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass

### DIFF
--- a/src/weather_model_graphs/save.py
+++ b/src/weather_model_graphs/save.py
@@ -9,8 +9,8 @@ from .networkx_utils import MissingEdgeAttributeError, split_graph_by_edge_attri
 
 try:
     import torch
-    import torch_geometric.utils.convert as pyg_convert
     import torch_geometric as pyg
+    import torch_geometric.utils.convert as pyg_convert
 
     HAS_PYG = True
 except ImportError:
@@ -18,12 +18,12 @@ except ImportError:
 
 
 def to_pyg(
-        graph: networkx.DiGraph,
-        output_directory: str,
-        name: str,
-        edge_features: List[str] | None = None,
-        node_features: List[str] | None = None,
-        list_from_attribute=None,
+    graph: networkx.DiGraph,
+    output_directory: str,
+    name: str,
+    edge_features: List[str] | None = None,
+    node_features: List[str] | None = None,
+    list_from_attribute=None,
 ):
     """
     Save the networkx graph to PyTorch Geometric format that matches what the
@@ -86,14 +86,16 @@ def to_pyg(
     def _get_edge_indecies(pyg_g):
         return pyg_g.edge_index
 
-    def _concat_pyg_features(pyg_g: "pyg.data.Data", features: List[str]) -> torch.Tensor:
+    def _concat_pyg_features(
+        pyg_g: "pyg.data.Data", features: List[str]
+    ) -> torch.Tensor:
         """Convert features from pyg.Data object to torch.Tensor.
-        Each feature should be column in the resulting 2D tensor (n_features, n_edges or n_nodes).
+        Each feature should be column in the resulting 2D tensor (n_edges or n_nodes, n_features).
         Note, this function can handle node AND edge features.
         """
         v_concat = []
         for f in features:
-            v = torch.tensor(pyg_g[f])  # make sure we have torch tensor
+            v = pyg_g[f]
             # Convert 1D features into 1xN tensor
             if v.ndim == 1:
                 v = v.unsqueeze(1)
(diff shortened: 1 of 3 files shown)


src/weather_model_graphs/save.py
```diff
@@ -59,6 +61,13 @@ def to_pyg(
             "install weather-mode-graphs[pytorch] to enable writing to torch files"
         )
 
+    # Default values for arguments
+    if edge_features is None:
+        edge_features = ["len", "vdiff"]
```
joeloskarsson, line 66: This makes sense as a default as it matches neural-lam (https://github.com/mllam/neural-lam/blob/a2ddcd4298d7865225c45866e7495664b3e382ab/neural_lam/create_mesh.py#L84) :+1:

src/weather_model_graphs/visualise/plot_2d.py
```diff
@@ -13,7 +13,7 @@ def nx_draw_with_pos(g, with_labels=False, **kwargs):
     if ax is None:
         _, ax = plt.subplots(figsize=(10, 10))
     networkx.draw_networkx(
-        ax=ax, G=g, pos=pos, hide_ticks=False, with_labels=with_labels, **kwargs
+        ax=ax, G=g, pos=pos, with_labels=with_labels, **kwargs
```
joeloskarsson, line 16: `hide_ticks` seems to be new from networkx version 3.3 (https://networkx.org/documentation/stable/reference/generated/networkx.drawing.nx_pylab.draw_networkx.html). Looks like this is the version we require https://github.com/mllam/weather-model-graphs/blob/2456064543200af1adfd17db0d1ccbb5f7441ca8/pyproject.toml#L11
so if you upgrade to 3.3 then this change should not be necessary.

src/weather_model_graphs/save.py
```diff
@@ -77,16 +86,20 @@ def to_pyg(
...
-            raise NotImplementedError(node_features_values)
-        return pyg_g.pos.to(torch.float32)
+    def _concat_pyg_features(pyg_g: "pyg.data.Data", features: List[str]) -> torch.Tensor:
+        """Convert features from pyg.Data object to torch.Tensor.
+        Each feature should be column in the resulting 2D tensor (n_features, n_edges or n_nodes).
+        Note, this function can handle node AND edge features.
+        """
+        v_concat = []
+        for f in features:
+            v = torch.tensor(pyg_g[f])  # make sure we have torch tensor
```
joeloskarsson, line 96: Just for my understanding, why is this necessary? Do you have a specific case where `pyg_g[f]` is not tensor already?

src/weather_model_graphs/save.py
```diff
@@ -77,16 +86,20 @@ def to_pyg(
...
-        # TODO: handle features of different types more generally, i.e. both single ("len") values and tuples (like "vdiff")
-        return torch.cat((pyg_g.len.unsqueeze(1), pyg_g.vdiff), dim=1).to(torch.float32)
-
-    def _get_node_features(pyg_g):
-        if node_features != ["pos"]:
-            raise NotImplementedError(node_features_values)
-        return pyg_g.pos.to(torch.float32)
+    def _concat_pyg_features(pyg_g: "pyg.data.Data", features: List[str]) -> torch.Tensor:
+        """Convert features from pyg.Data object to torch.Tensor.
+        Each feature should be column in the resulting 2D tensor (n_features, n_edges or n_nodes).
```
joeloskarsson, line 91:
```suggestion
        Each feature should be column in the resulting 2D tensor (n_edges or n_nodes, n_features).
```

joeloskarsson (changes requested):
This is great, thanks a lot for this contribution! :smile: 

I went over the code and left some small comments. I don't think there is any need for adding additional tests here as 
1. the saving functionality with default feature arguments is already tested in https://github.com/mllam/weather-model-graphs/blob/2456064543200af1adfd17db0d1ccbb5f7441ca8/tests/test_save.py#L18 
2. we don't have any good functionality to add additional edge/node features that is now saveable with this more universal approach and could be used in a test (as far as I remember, let me know if I am wrong)

There are a couple small additional things that would be good to do before merging this:

1. The linting seems to fail due to the code formatting. This should be easily fixed by just running the pre-commit hooks once. 
2. Go ahead and add an entry to the changelog! :slightly_smiling_face:

src/weather_model_graphs/visualise/plot_2d.py
maxiimilian: Oh, I should have checked that. I will revert the change.

src/weather_model_graphs/save.py
maxiimilian: No, good point. The data from `pyg.data.Data` should always be tensors. I will remove it.

joeloskarsson (approved):
Everything good now, I'll merge this in. Thanks again for this contribution!

PR #26, opened 2024-09-26
Fix wrong number of mesh levels when grid is multiple of refinement factor

joeloskarsson:
## Describe your changes

There is currently a bug where you can not make hierarchical graphs if the grid dimensions are an exact multiple of the refinement factor. This is best illustrated with an example: 

* If you have a 27 x 27 grid
* grid_refinement_factor=3 and level_refinement_factor=3

Then you should be able to have a first mesh level of 9x9 and a second of 3x3 mesh nodes. This does currently not work, as the code return that the maximum number of possible mesh levels is 1, rather than 2.

We tracked down this issue to the casting to int in 

https://github.com/mllam/weather-model-graphs/blob/1d1407b199e7919c6b8c9bd1f355c4292246eb23/src/weather_model_graphs/create/mesh/mesh.py#L116-L121

Due to the numerical precision of this calculation it sometimes ends up with something like 1.999999999999996 rather than exactly 2. This then gets floored to 1.

This PR fixes this by adding a tiny epsilon before doing this rounding. It also adds a test for this.

## Type of change

- [x] 🐛 Bug fix (non-breaking change that fixes an issue)
- [ ] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] ~I have updated the documentation to cover introduced code changes~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [x] I have requested a reviewer and an assignee (assignee is responsible for merging)

## Checklist for reviewers

Each PR comes with its own improvements and flaws. The reviewer should check the following:
- [x] the code is readable
- [x] the code is well tested
- [x] the code is documented (including return types and parameters)
- [x] the code is easy to maintain

## Author checklist after completed review

- [x] I have added a line to the CHANGELOG describing this change, in a section
[...]

PR #21, opened 2024-09-10
Create different number of mesh nodes in x- and y-direction

joeloskarsson:
## Describe your changes

Currently the 2d multirange graphs created for multiscale and hierarchical graphs are restricted to laying  out mesh nodes in an $n \times n$ grid. This is suitable for forecasting areas that are roughly square, but a poor choice for rectangular areas with a large difference between it's height and width.

This PR updates the way multiscale and hierarchical graphs are created to compute the refinement (or equivalently the number of mesh nodes at each level) separately in the x- and y-directions. The result of this is probably best understood by an example:

Here a mesh graph is created for an area with $10 \times 30$ grid nodes. We are making a hierarchical graph, plotting only nodes. Grid nodes have level -1. Before this change the graph would look like this:

hi_old

Note that for level 0 there are 8 mesh nodes created in each direction, although this is way too many for the y-direction. This results in very short distances within the mesh in the y-direction compared to the x-direction. The higher levels share the same problem.

After this change the same graph looks like this:

hi_new

Now the distance between mesh nodes is roughly the same in the x- and y-direction. Note that with this change we can no longer fit 3 mesh levels. This is actually desirable behavior, as the height of this area is actually too small to properly fit 3 levels.

This change builds upon #19 , and should be merged after that.

## Issue Link

This provides some potential answers to #2. If those outstanding questions do not require further discussion I think this can close #2.

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
- [x] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [x] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [x] I have updated the documentation to cover introduced code changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).
- [x] I have requested a reviewer and an assignee (assignee is responsible for merging)

## Checklist for reviewers

[...]

PR #32, opened 2024-10-13
Allow non-regularly gridded and lat-lon coordinates 

### LINKED ISSUE #29: Allow non-regularly gridded and lat-lon coordinates
This issue is to document and discuss some enhancements I am looking into w.r.t. the coordinates used as inputs. 

## The current situation
Currently the grid node coordinates fed to wmg are a numpy array of shape [2, N_y, N_x]. This assumes that the grid nodes are on a regular grid. It is also the case that these are assumed to be cartesian coordinates, in the projection used for the LAM data.

## Generalizing to non-regular gridded data
My idea to generalize this is to change so that the coordinate array `xy` instead has shape [N_grid_nodes, 2] with e.g. `xy[:,0]` containing the x-coordinate for all grid nodes. Existing mesh creation algorithms would still have to work with a rectangular area, for which they can use the axis-aligned bounding box of the grid nodes. Other, new, mesh creation methods could use the convex hull of the set of grid nodes.

## Generalizing to allow lat-lon coordinates
For lat-lon coordinates the same [N_grid_nodes, 2] shape could be used for the coordinates. It would also be good to feed a projection specification (I propose a Cartopy projection instance, https://scitools.org.uk/cartopy/docs/latest/reference/generated/cartopy.crs.Projection.html). Using the projection the lat-lon coordinates can then be mapped to a euclidean space and all our current methods can be applied as is. Having the option of lat-lon coordinates however also opens up the opportunity for mesh creation methods that work directly with these.

A question is how to tell wmg if in-projection or lat-lon coordinates are given. Some options are:

* Have separate arguments `xy` and `lat_lon` where only one has to be given. 
* Have only one argument `coords`, and then a separate boolean flag  e.g. `uses_lat_lon`.
* One could also enforce that a projection has to be given when using lat-lon coordinates, and use that implicitly as the `uses_lat_lon` flag above.
* Another option is really to only allow lat-lon coordinates from here onwards.

Not really sure what is the best option of the ones above. Any input is appreciated.

joeloskarsson:
## Describe your changes

This PR allows for more flexibility in the coordinates used to create graphs. They do no longer need to be on a regular $N_x \times N_y$ grid, but can be at arbitrary positions. This means that the main coordinate input no longer has the shape `[2, Ny, Nx]`, but instead `[N_grid_points, 2]` with each column representing a coordinate for all grid points. This makes it easy to also allow for lat-lon input coordinates, and use a specified projection to map these to a euclidean space.

To get an overview of the lat-lon functionality I recommend checking out the new page I added to the documentation.

The motivation for this change is to allow more flexibility in the regions graphs are built for. This is in particular useful when we consider grid points as the union of some LAM-region and a boundary of global grid-points, possibly on different gridding. With this change such cases can be handled without issues.

There are a few noteworthy changes that comes with this feature:

* I broke out the flat graph creation into a function `create_flat_singlescale_mesh_graph`, as it made more sense that also this type of mesh graph had its own function. This made sense to do here as the amount of related code was growing.
* I also moved some utility functions around the testing (for creating dummy coordinate arrays) into their own `tests/utils.py` file. This made sense as these functions were duplicated in multiple files before and I had to change them with this new coordinate setup.
* The argument `grid_refinement_factor` was earlier based on the number of grid nodes in x- and y-directions. As we do no longer assume that these grid nodes lay in regular rows and columns this definition will no longer work. Instead we will have to define the grid refinement in terms of "mesh nodes / distance", using the actual coordinate values. The argument is also being renamed to `mesh_node_distance` to match this.
* In the high-level functions that are used to create graphs the `xy` argument is renamed to `coords`, as this can be lat-lons or euclidean x- and y-coordinates. If a projection is given this is assumed to be lat-lons, otherwise euclidean. For functions further down (e.g. `create_single_level_2d_mesh_graph`), that still take only euclidean coordinates, this argument is still called `xy`.

### New dependencies

This PR introduces a dependency to cartopy, as that is needed for the projection specification when working with lat-lons.

### Merge and release planning

* ~~It would be good if this change went in after #26 as both deal with the refinement factor calculations.~~ (Now merged)
* We might want to create a new release before merging this, as this is a breaking change that will alter how the main input to the package is formatted. It could be nice to separate all previous changes in a release before adding this on.

## Issue Link

This PR closes #29 

## Type of change

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist for reviewers

- [x] the code is readable
- [x] the code is well tested

## Checklist for assignee

- [x] the tests pass


joeloskarsson:
This is now ready for review :partying_face:

sadamov (changes requested):
Thanks a lot @joeloskarsson these new graphs look real nice. Expecially for domains closer to the poles. But I also generated a test_plot close to the equator which looks well designed to me:
image

Some comments:
- In `Changelog.md` also mention that `mesh_node_distance` is the new name of `grid_refinement_factor`.
- Is `Euclidean` coordinates or `Cartesian` coordinates easier to understand? Maybe Euclidean is more accurate regardless...?
- These lat-lon based graphs would benefit a lot from heterogeneous graph design it seems (i.e. not living on a equidistant grid of nodes). This is a topic for a future PR though, right?

src/weather_model_graphs/create/base.py
```diff
@@ -53,41 +58,52 @@ def create_all_graph_components(
...
     """
     graph_components: dict[networkx.DiGraph] = {}
 
-    if len(xy.shape) != 3:
-        raise NotImplementedError(
-            "Mesh coordinates are assumed to lie on a regular grid so that "
-            "the coordinates values are given with an array of shape [2, nx, ny]"
+    assert (
+        len(coords.shape) == 2 and coords.shape[1] == 2
+    ), "Grid node coordinates should be given as an array of shape [2, num_grid_nodes]."
```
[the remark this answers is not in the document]
joeloskarsson: Ah yes, well spotted typo!

joeloskarsson:
>  In Changelog.md also mention that mesh_node_distance is the new name of grid_refinement_factor.

Yes, that's a good idea. I haven't added any changelog entry yet since I think that we should probably create a new release before merging this, and then the changelog will be updated anyhow and this entry go as the first under `unreleased`.

>  Is Euclidean coordinates or Cartesian coordinates easier to understand? Maybe Euclidean is more accurate regardless...?

Thinking a bit more about this (and reminding myself about the exact definitions), I actually think that we should use Cartesian rather than Euclidean everywhere. My reasoning is that cartesian refers to a coordinate system, whereas euclidean refers to a type of geometric space (with arbitrary coordinate system, e.g. could be spherical coordinates). Now, the surface of the earth is not a euclidean space and lat-long not cartesian coordinates, so there's no real risk of substantial misunderstandings, but I guess "Cartesian coordinates" makes sense whereas "Euclidean coordinates" is technically ambiguous. 

@sadamov If you agree with this reasoning I would go ahead and change all comments and names to use Cartesian :smile: 

> These lat-lon based graphs would benefit a lot from heterogeneous graph design it seems (i.e. not living on a equidistant grid of nodes). This is a topic for a future PR though, right?

Yes, they would definitely benefit a lot from that. But I agree that that's for a future PR. One simple thing I was trying out was to just remove any mesh nodes that does not have any grid node  connections. One could do that for the flat Keisler-like graphs, but I realized that it becomes tricky for the multiscale(?) and hierarchical graphs. In order to now have different behavior for the different archetypes I left that out now. Better to revisit and think a bit more about when it is clearly needed.

sadamov (approved):
> Cartesian rather than Euclidean everywhere. My reasoning is that cartesian refers to a coordinate system, whereas euclidean refers to a type of geometric space (with arbitrary coordinate system, e.g. could be spherical coordinates). 

Yes I agree, this was my main concern when I wrote the comment. Cartesian coordinates :+1: 

> In order to now have different behavior for the different archetypes I left that out now. Better to revisit and think a bit more about when it is clearly needed.

Makes sense to me!

Still unsure about our merging policies... but from my side ready to merge into main.

joeloskarsson:
Note: This PR should have a review by @leifdenby , as by the roadmap. Then it is ready to merge.

src/weather_model_graphs/create/base.py
```diff
@@ -70,27 +75,38 @@ def create_all_graph_components(
...
+            "No `projection` given: Assuming `coords` contains in-projection Cartesian coordinates."
+        )
+        xy = coords
+    else:
+        logger.debug(
+            f"`projection` Proj({projection}) given, `coords` treated as lat-lons."
+        )
+        # Convert lat-lon coords to Cartesian xy
+        xyz = projection.transform_points(
+            src_crs=ccrs.PlateCarree(), x=coords[:, 0], y=coords[:, 1]
```
observingClouds, line 100: Do we need the projections to be cartopy projections or would it suffice to have pyproj projections? Having done a deep dive into cartopy in https://github.com/mllam/mllam-data-prep/issues/33 I think it would be great to keep cartopy out of the loop as long as possible, e.g. only use it for plotting. I think this transformation should also be possible with:
```python
from pyproj import Transformer

transformer = Transformer.from_crs("EPSG:4326", "ESRI:54001")
x, y = transformer.transform(12.4924, 41.8902)
```
The transformer also handles cartopy.crs.CRS if provided, so this approach would just be more general.

joeloskarsson: I think it should be no problem to rely on pyproj over cartopy for this, we just need a way to transform the coordinates. I only want to make sure it is easy for users to specify the projection. Could you give an example of how an input for `projection` at https://github.com/joeloskarsson/weather-model-graphs/blob/4fb69dcf2a91560e5dd80247dd313ec7d88f23c6/src/weather_model_graphs/create/base.py#L40 would look like with pyproj?

I guess it would be a `pyproj.Proj`? Is there a list similar to https://scitools.org.uk/cartopy/docs/v0.15/crs/projections.html for pyproj that we could refer users to for how to construct them?

leifdenby: This is a good point @observingClouds, but I agree with @joeloskarsson that allowing people to provide projections as `cartopy.crs.Projection` objects is something people will be familiar with. Could we support both? Or is that going to be too complicated do you think @observingClouds?

If you were to pass in the projection as an argument to `create_all_graph_components(..., crs=...)` what would that look like in this example?

The other thing to consider is that `weather-model-graphs` is built to produce visualisations, but I had made that optional (and therefore put `matplotlib` in the `[visualisation]` optional dependency group in `pyproject.toml` for this reason: https://github.com/mllam/weather-model-graphs/blob/main/pyproject.toml#L23. Maybe if we work out how to support both methods of defining the projection then `cartopy` can go into the `visualisation` dependency group?

leifdenby:
I'm in the process of giving this a thorough read through and executing the notebooks etc as I go through. In general this is really excellent (as usual), and I fully agree with all of the following three changes:

> * I broke out the flat graph creation into a function `create_flat_singlescale_mesh_graph`, as it made more sense that also this type of mesh graph had its own function. This made sense to do here as the amount of related code was growing.
> 
> * I also moved some utility functions around the testing (for creating dummy coordinate arrays) into their own `tests/utils.py` file. This made sense as these functions were duplicated in multiple files before and I had to change them with this new coordinate setup.
> 
> * The argument `grid_refinement_factor` was earlier based on the number of grid nodes in x- and y-directions. As we do no longer assume that these grid nodes lay in regular rows and columns this definition will no longer work. Instead we will have to define the grid refinement in terms of "mesh nodes / distance", using the actual coordinate values. The argument is also being renamed to `mesh_node_distance` to match this.

The following I also really like, but I have a question/suggestion wrt it:

> * In the high-level functions that are used to create graphs the `xy` argument is renamed to `coords`, as this can be lat-lons or euclidean x- and y-coordinates. If a projection is given this is assumed to be lat-lons, otherwise euclidean. For functions further down (e.g. `create_single_level_2d_mesh_graph`), that still take only euclidean coordinates, this argument is still called `xy`.

As I am reading it the intended use in your new implementation has two main use cases (in terms of how the coordinates are defined):

```python
# 1. Using the coordinate values as-is, measuring distance in units of
# coordinate values with the euclidean distance metric. Relevant when working
# with cartesian coordinates.
graph = wmg.create.archetype.create_keisler_graph(
    coords,
    mesh_node_distance=10,  # distance in units of coordinate values, measured as euclidean distance
    projection=None,
)

# 2. First projecting the coordinates to a different coordinate system,
# measuring distance in units of the projection with the euclidean distance
# metric. Relevant when working with lat/lon coordinates.
graph = wmg.create.archetype.create_keisler_graph(
    coords,
    mesh_node_distance=1.0e6,  # distance in units of projected coordinates, measured as euclidean distance
    projection=ae_projection
)
```

with the current function signature being:

```python
def create_keisler_graph(
    coords: np.ndarray,
    mesh_node_distance: float,
    projection: Optional[pyproj.Proj] = None,
) -> nx.Graph:
```

Where it is assumed that distance is measured with the euclidean distance metric, with either the coordinate values as-is or first projecting.

I think it would be cool to extend this in future by allowing for a different distance metric to be used. That would allow us to create graphs from lat/lon coordinates directly (and thereby have graphs that fully wrap the Earth).

For example we could then do:

```python
# 1. Using the coordinate values as-is, measuring distance in units of
# coordinate values with the euclidean distance metric. This would be useful
# for creating graphs from cartesian coordinates.
graph = wmg.create.archetype.create_keisler_graph(
    coords,
    mesh_node_distance=10,
    projection=None,
    distance_metric="euclidean"
)

# 2. First projecting the coordinates to a different coordinate system,
# measuring distance in units of the projection with the euclidean distance
# metric. This would be useful for creating graphs from lat/lons for limited
# areas.
graph = wmg.create.archetype.create_keisler_graph(
    coords,
    mesh_node_distance=mesh_distance,
    projection=ae_projection,
    distance_metric="euclidean"
)

# 3. Using the coordinate values as-is, measuring distance in great circle
# distance with the haversine distance metric. This would be useful for
# creating graphs from lat/lon coordinates directly. We should default to the
# distance being in SI units (as in meters) and so use a fixed Earth radius in
# meters.
graph = wmg.create.archetype.create_keisler_graph(
    coords,
    mesh_node_distance=1.0e4,  # distance of 10 km between coordinates given in lat/lon
    projection=None,
    distance_metric="haversine"
)
```

With the function signature changed to:

```python
def create_keisler_graph(
    coords: np.ndarray,
    mesh_node_distance: float,
    projection: Optional[pyproj.Proj] = None,
    distance_metric: str = "euclidean"
) -> nx.Graph:
```

So that by default the function will use euclidean distance with the coordinate values, but the metric can be changed to haversine (which gives approximate great circle distances in meters between lat/lon coordinates, assuming a fixed Earth radius).

All that is then required is to pass the `distance_metric` argument down the function calls and where the actual distance is calculated and used implement the logic for the different distance metrics.

What do you think of this idea? If it sounds good to you then maybe we could go ahead and add `distance_metric` as an argument to the function (with default value `euclidean`) to make the metric used explicit, and then that would make it easy to add the haversine distance metric in the future.

Of course one could argue that it should be possible to provide the Earth radius, but maybe that could become an optional `metric_argument` down the line. Or we allow people to optionally provide a projection that makes it clear that the coordinates are lat/lon and in that projection object they can then set the globe radius (e.g. with `cartopy.crs.PlateCarree(globe=cartopy.crs.Globe(semimajor_axis=..., semiminor_axis=...))` or something like that)

src/weather_model_graphs/create/base.py
observingClouds: `Transformer.from_crs()` accepts also cartopy.crs.CRS derived types, so only the function here needs to be changed and the examples can remain the same from what I can see:

```python
from pyproj import Transformer
import cartopy.crs as ccrs
import numpy as np

coords = np.array([[12.4924],[41.8902]])

ccrs.Mercator().transform_points(ccrs.PlateCarree(), x=coords[0], y=coords[1])
# array([[1390647.6067859 , 5116008.22708512,       0.        ]])

Transformer.from_crs(ccrs.PlateCarree(), ccrs.Mercator(), always_xy=True).transform(coords[0], coords[1])
#(array([1390647.6067859]), array([5116008.22708512]))

Transformer.from_crs("EPSG:4326", "EPSG:3857", always_xy=True).transform(coords[0], coords[1])
# (array([1390647.60678589]), array([5144546.10036227])) # slight changes as the standards are differently defined than the general cartopy objects
```

observingClouds: @joeloskarsson pyproj does not have a list of available general projections, they are more flexible.

observingClouds: > I guess it would be a pyproj.Proj
I have to say I'm not an expert on projections, but if I read https://github.com/joeloskarsson/weather-model-graphs/blob/4fb69dcf2a91560e5dd80247dd313ec7d88f23c6/src/weather_model_graphs/create/base.py#L40 correctly, then a coordinate reference system (CRS) is requested and not a projection as the input is projected to Plate Carree, opposed to just calling a projection that would have src and target CRS already defined.

So maybe `crs` instead of `projection` would be a better keyword argument. But the difference is small, particularly because also _projected coordinate reference systems_ exist. In the current form of the function call, the input would be of type pyproj.crs.CRS or cartopy.crs.CRS (as it is already).

Does this make sense @joeloskarsson ?

leifdenby: > So maybe `crs` instead of `projection` would be a better keyword argument. But the difference is small, particularly because also _projected coordinate reference systems_ exist

I think this makes sense. I was just reading up on what exactly at CRS is to make sure I am using the nomenclature correctly (https://docs.qgis.org/3.34/en/docs/gentle_gis_introduction/coordinate_reference_systems.html#coordinate-reference-system-crs-in-detail). I think it would be clearer to say that the argument defines the coordinate reference system, whether that is not projected (i.e. describes positions on the Earth for example using lat/lons and specific Earth properties) or projected (i.e. builds on a not projected coordinate reference system by applying a specific projection)

docs/creating_the_graph.ipynb
leifdenby, line 1: The text in the section `# The mesh nodes and graph` probably should also be changed as it the moment it reads:

> Lets start with a simple mesh which only has nearest neighbour connections. At the moment `weather-model-graphs` _creates a square mesh that sits within the spatial domain spanned by the grid nodes. Techniques for adding non-square meshes are in development._

I added the emphasis. That isn't quite true anymore, right?

src/weather_model_graphs/create/archetype.py
leifdenby, line 1: I wanted to check that the `mest_node_distance` arguments works as I would expect, so I created a set of grid points spanning `[0,20]^2` and made a Keisler graph with `mesh_node_distance=5.0`, looks good to me by eye (I would say the those mesh nodes are `5.0` apart, and centered on the domain too) - nice!

```python
import numpy as np
import matplotlib.pyplot as plt
import weather_model_graphs as wmg

def _create_fake_xy(N=10):
    x = np.linspace(0.0, N, N)
    y = np.linspace(0.0, N, N)
    xy_mesh = np.meshgrid(x, y)
    xy = np.stack([mg_coord.flatten() for mg_coord in xy_mesh], axis=1)  # Shaped (N, 2)
    return xy

xy = _create_fake_xy(20)
graph = wmg.create.archetype.create_keisler_graph(coords=xy, mesh_node_distance=5.0)
wmg.visualise.nx_draw_with_pos_and_attr(
    graph, node_size=30, edge_color_attr="component", node_color_attr="type"
)
```

billede

leifdenby (commented):
Looks really great to me. There isn't that you've written that I think is incorrect, I've only made some minor suggestions for small improvements. And then there's the suggestion to make the distance metric used explicit so that we can later introduce other distance metrics (for example haversine)

src/weather_model_graphs/create/mesh/mesh.py
```diff
@@ -94,36 +94,34 @@ def create_multirange_2d_mesh_graphs(
...
-
-    # Need to add a small epsilon before flooring to int, due to numerical
-    # issues with the computation above
-    eps = 1e-8
-    max_mesh_levels = (max_mesh_levels_float + eps).astype(int)  # (2,)
-    nleaf = grid_refinement_factor * (
-        level_refinement_factor**max_mesh_levels
-    )  # leaves at the bottom in each direction, if using max_mesh_levels
+    # max_mesh_coord/(level_refinement_factor^mesh_levels) = 1
+    max_mesh_levels_float = np.log(extent_nodes_bottom_mesh) / np.log(
```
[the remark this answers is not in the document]
joeloskarsson: Hmm, I think it is correct. With $m$ = max_mesh _coord, $r$ = level_refinement_factor and $l$ = mesh_levels (all positive):

$$
\frac{m}{r^l} = 1 \Leftrightarrow \log(m) - l \log(r) = 0 \Leftrightarrow l = \frac{\log(m)}{\log(r)}
$$

What is a bit confusing though is that the variable names are not the same as in the comment, so that could be improved.

docs/creating_the_graph.ipynb
joeloskarsson: I think this is still mostly true. This PR will still only create square meshes (in the sense of mesh graph). Or maybe better described as "rectangular meshes", as the rows and columns of mesh nodes are not necessarily equally spaced.  "spanned by the grid nodes" is not very precise wording (it's rather within the axis-aligned bounding box of the grid nodes), but gets the message across.

src/weather_model_graphs/create/base.py
joeloskarsson: Thanks for explaining and good suggestions! 

> then a coordinate reference system (CRS) is requested and not a projection as the input is projected to Plate Carree, opposed to just calling a projection that would have src and target CRS already defined.

This is a very good point. I think I have been very sloppy with the differences between crs and projection both when talking about this and thinking about it :sweat_smile: It is absolutely true that the argument now is a crs and, when given, the coords input is always transformed from PlateCarree to this crs.  

I think it is important to allow for cartopy.crs.CRS as options, as that is a lot easier for most users. But as pyproj Transformer.from_crs() accepts also that I think it would be great to change to using that. Then the pyproj only gives additional flexibility which is great!

As for the keyword name I agree that crs would be more accurate then projection. It seems however a bit confusing to me that the logic would be 1. If crs == None: assume that coordinates are already given in the desired coordinate system for graph creation 2. if crs != None: Assume coordinates are lat-lons and transform them to the given crs.
I propose to instead have two keyword arguments: coords_crs and graph_crs, where they both default to None (or maybe coords_src defaults to Plate Carree?). When both are given the coords are transformed from coords_crs to the graph_crs, where the graph creation is carried out. That also adds even more flexibility, as it allows for transforming from arbitrary crs:s, not just lat-lons.

observingClouds: > I propose to instead have two keyword arguments: coords_crs and graph_crs

I like this suggestion as it is more explicit and adds flexibility.

docs/creating_the_graph.ipynb
leifdenby: ah right, yes I should made it clearer that I meant we should say *rectangular* rather than *square* since you added that functionality in https://github.com/mllam/weather-model-graphs/pull/21

src/weather_model_graphs/create/mesh/mesh.py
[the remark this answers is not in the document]
leifdenby: ok, yes, that was why I couldn't get it to line up. But the maths checks out! Thank you for writing that

tests/test_graph_creation.py
```diff
@@ -113,33 +99,65 @@ def test_create_rectangular_graph(kind):
...
+def test_create_irregular_grid(kind):
+    """
+    Tests that graphs can be created for irregular layouts of grid points
+    """
+    xy = test_utils.create_fake_irregular_coords(100)
+    fn_name = f"create_{kind}_graph"
+    fn = getattr(wmg.create.archetype, fn_name)
+
+    # ~= 20 mesh nodes in bottom layer in each direction
+    fn(coords=xy, mesh_node_distance=0.05)
```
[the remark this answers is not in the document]
joeloskarsson: https://github.com/mllam/weather-model-graphs/pull/32/commits/05a0cc6e066ae952b4e07069d92c1fa006d7bad5

docs/creating_the_graph.ipynb
joeloskarsson: https://github.com/mllam/weather-model-graphs/pull/32/commits/339feff2ef5ef081fa73600536c232e17a834263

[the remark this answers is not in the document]
joeloskarsson: weather-model-graphs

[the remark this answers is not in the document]
joeloskarsson: I tested locally that this way of linking to another notebook works in jupyter-book.

pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "loguru>=0.7.2",
     "networkx>=3.3",
     "scipy>=1.13.0",
+    "cartopy>=0.24.1",
```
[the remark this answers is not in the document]
joeloskarsson: Ok had to bring this back to the visualisation group for the docs and tests (https://github.com/mllam/weather-model-graphs/pull/32/commits/8e3c1cbf45b7ed758eb81550c150628f76bc2370)

src/weather_model_graphs/create/base.py
observingClouds: Thanks @joeloskarsson for the changes! LGTM!

docs/lat_lons.ipynb
leifdenby, line 1: This is really nicely written! All very clear and the code executes perfectly too. My only comment is whether you meant to write `CRS:s` rather than `CRSs`? I'm not familiar with adding `:` to pluralise, but we can keep it as is if you prefer :)

src/weather_model_graphs/create/archetype.py
```diff
@@ -18,20 +23,24 @@ def create_keisler_graph(xy_grid, grid_refinement_factor=3):
...
-    xy_grid: np.ndarray
-        2D array of grid point positions.
-    grid_refinement_factor: float
-        Refinement factor between grid points and mesh
+    coords: np.ndarray
+        2D array of grid point positions, either in-projection Cartesian coordinates or lat-lons
+    mesh_node_distance: float
+        Distance (in x- and y-direction) between created mesh nodes,
+        in coordinate system of coords
+    projection: cartopy.crs.CRS or None
```
leifdenby, line 31: I think this docstring line needs updating to reflect that `coords_crs` and `graph_crs` are now provided separately.

src/weather_model_graphs/create/archetype.py
```diff
@@ -61,26 +77,30 @@ def create_graphcast_graph(
...
+        2D array of grid point positions, either in-projection Cartesian coordinates or lat-lons
+    mesh_node_distance: float
+        Distance (in x- and y-direction) between created mesh nodes,
+        in coordinate system of coords
     level_refinement_factor: int
         Refinement factor between grid points and bottom level of mesh hierarchy
         NOTE: Must be an odd integer >1 to create proper multiscale graph
     max_num_levels: int
         The number of levels of longer-range connections in the mesh graph.
+    projection: cartopy.crs.CRS or None
```
leifdenby, line 90: I think this docstring line needs updating to reflect that `coords_crs` and `graph_crs` are now provided separately.

src/weather_model_graphs/create/archetype.py
```diff
@@ -119,23 +146,27 @@ def create_oskarsson_hierarchical_graph(
...
-    grid_refinement_factor: float
-        Refinement factor between grid points and bottom level of mesh hierarchy
+    coords: np.ndarray
+        2D array of grid point positions, either in-projection Cartesian coordinates or lat-lons
+    mesh_node_distance: float
+        Distance (in x- and y-direction) between created mesh nodes in bottom level,
+        in coordinate system of coords
     level_refinement_factor: float
         Refinement factor between grid points and bottom level of mesh hierarchy
+    projection: cartopy.crs.CRS or None
```
leifdenby, line 156: And there, I think this docstring line needs updating to reflect that `coords_crs` and `graph_crs` are now provided separately.

src/weather_model_graphs/create/base.py
```diff
@@ -70,27 +76,48 @@ def create_all_graph_components(
     - "containing_rectangle": For each grid node, find the rectangle with 4 mesh nodes as corners
         such that the grid node is contained within it. Connect these 4 (or less along edges)
         mesh nodes to the grid node.
+
+    `coords_crs` and `graph_crs` should either be a pyproj.crs.CRS or None.
+    Note that this includes a cartopy.crs.CRS. If both are given the coordinates
+    will be transformed from their original Coordinate Reference System (`coords_crs`)
+    to the CRS where the graph creation should take place (`graph_crs`).
+    If any one of them is None the graph creation is corried out using the original coords.
```
leifdenby, line 84:
```suggestion
    If any one of them is None the graph creation is carried out using the original coords.
```

leifdenby (commented):
Only some very minor comments/suggestions left. Great!

src/weather_model_graphs/create/base.py
leifdenby: Looks good to me too @joeloskarsson. I'll resolve this conversation :)

pyproject.toml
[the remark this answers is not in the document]
leifdenby: great, thanks!

tests/test_graph_creation.py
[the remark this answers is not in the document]
leifdenby: THANK YOU!

docs/lat_lons.ipynb
joeloskarsson: I could swear that :s plural is a thing for abbreviations, but googling it I can't find anything, so somehow I have just dreamt this up :smile: Anyhow, CRSs makes more sense, I can change that.

src/weather_model_graphs/create/archetype.py
joeloskarsson: Right, forgot to change those docstrings as well. Thanks!

src/weather_model_graphs/create/mesh/mesh.py
```diff
@@ -139,9 +137,7 @@ def create_multirange_2d_mesh_graphs(
     G_all_levels = []
     for lev in range(mesh_levels_to_create):  # 0-index mesh levels
         # Compute number of nodes on level separate for each direction
-        nodes_x, nodes_y = (
-            nleaf / (grid_refinement_factor * (level_refinement_factor**lev))
-        ).astype(int)
+        nodes_x, nodes_y = (nleaf / (level_refinement_factor**lev)).astype(int)
```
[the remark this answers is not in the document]
joeloskarsson: Hmm, not obvious to me that that is desirable. There's two possible behaviors in terms of how many levels can be created: We stop making new levels when one of the directions (x or y) can't be refined further, or we stop making new levels when none of the directions can be refined further. Not obvious which one is preferable to me, but I don't see myself wanting to create graphs following the second choice. One could make this an option. Or just go with the second option and then you have to use the max level argument to get rid of the top levels you don't want.

Anyway I think that a change to this would require changing the number of level computations above as well, for that to be correct. It should probably also be combined with separate level_refinement_factors for x- and y-directions.

My preference is to leave this as is in this PR, and can be tweaked later.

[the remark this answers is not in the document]
leifdenby: Ok, good points. I agree. Thanks for typing this out. Let leave as it is for now and people can adjust if they want to

tests/test_graph_creation.py
```diff
@@ -67,12 +53,12 @@ def test_create_graph_archetype(kind):
 M2M_CONNECTIVITY_OPTIONS = dict(
     flat=[],
     flat_multiscale=[
-        dict(max_num_levels=3, grid_refinement_factor=3, level_refinement_factor=3),
-        dict(max_num_levels=1, grid_refinement_factor=5, level_refinement_factor=5),
+        dict(max_num_levels=3, mesh_node_distance=3, level_refinement_factor=3),
```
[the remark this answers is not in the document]
leifdenby: we don't need to, let's leave as is. I was just curious :)

joeloskarsson:
Should be all fixed now @leifdenby :)

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- Add coords_crs and graph_crs arguments to allow for using lat-lons coordinates
+  or other CRSs as input. These are then converted to the specific CRS used when
+  constructing the graph.
+  [\#32](https://github.com/mllam/weather-model-graphs/pull/32), @joeloskarsson
+
+### Changed
+
+- Change coordinate input to array of shape [N_grid_points, 2], to allow for
+  non-regularly gridded coordinates
```
leifdenby, line 20:
```suggestion
- Change coordinate input to array of shape [N_grid_points, 2] (was previously 
  [2, Ny, Nx]), to allow for non-regularly gridded coordinates
```

leifdenby (commented):
Looks great, just minor tweak suggested to make the change to coordinate shape explicit in CHANGELOG

leifdenby (approved):

PR #34, opened 2024-10-21
Add a decoding mask option to only include subset of grid nodes in m2g

joeloskarsson:
## Describe your changes
For LAM forecasting it is reasonable to not always predict values in nodes that are considered part of the boundary forcing. This is in particular a change planned for neural-lam. When we consider the graphs involved, this means that the g2m edges should only connect to a subset of the grid nodes. 

This PR introduces an option `decode_mask`  that allows for specifying an Iterable of booleans (e.g. a numpy-array) specifying which of the grid nodes should be included in the decoding-part of the graph (m2g). This allows in the LAM case to specify such a mask with True for the inner region nodes.

This builds on #32, which should be merged first. Here is a diff for only this PR in the meantime: https://github.com/joeloskarsson/weather-model-graphs/compare/general_coordinates...decoding_mask

## Type of change

- [ ] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


sadamov (approved):
Looks good! Agree that this is required. Testing for "less than" seems sufficient :+1: 

More general comment: I am slightly worried about the mask being generated in `mllam-data-prep` and then being used here and in `neural-lam`. We probably need some solid checks in `neural-lam` about the dimensions of the graph, the data-tensors and the mask.

joeloskarsson:
I agree that there is a need for checks w.r.t. the dimensions of the graph and feature tensors. The work I started on https://github.com/mllam/neural-lam/tree/graph_classes does contain quite a lot of this that I think will be useful, e.g. 

- https://github.com/mllam/neural-lam/blob/c097c582a3329c405ffc0879e7c27582d6735ab1/neural_lam/graphs/base_weather_graph.py#L101-L118
- https://github.com/mllam/neural-lam/blob/c097c582a3329c405ffc0879e7c27582d6735ab1/neural_lam/graphs/flat_weather_graph.py#L35-L49.

joeloskarsson:
This PR is now mainly waiting on #32 to be merged

joeloskarsson:
To get this to actually work with neural-lam (or in particular with https://github.com/mllam/neural-lam/pull/84) I had to introduce some sorting of nodes and sub-graphs after splitting. Otherwise I was getting arbitrary node indices. 

@sadamov @leifdenby could you please have a look at https://github.com/mllam/weather-model-graphs/pull/34/commits/350e0c0b8166594f57d20aabc39c92d59d72b82b and check that you don't see any problem with this.

sadamov:
@joeloskarsson to me this restriction seems justified. It's also an easy to understand requirement for folks who want to create their graphs outside of WMG. Just make sure to add it to both README's here and in NL.
(and thanks for your work on this! :handshake:)

src/weather_model_graphs/save.py
```diff
@@ -106,11 +106,14 @@ def _concat_pyg_features(
...
-                ).values()
-            )
+            sub_graphs = [
+                kv[1]
+                for kv in sorted(
+                    split_graph_by_edge_attribute(
+                        graph=graph, attr=list_from_attribute
+                    ).items()
+                )
+            ]
```
SimonKamuk, line 116: Minor change, but in my opinion better readability
```suggestion
            sub_graphs = [
                value
                for key, value in sorted(
                    split_graph_by_edge_attribute(
                        graph=graph, attr=list_from_attribute
                    ).items()
                )
            ]
```

src/weather_model_graphs/create/archetype.py
```diff
@@ -18,20 +20,29 @@ def create_keisler_graph(xy_grid, grid_refinement_factor=3):
...
+        Distance (in x- and y-direction) between created mesh nodes,
+        in coordinate system of coords
+    projection: cartopy.crs.CRS or None
+        Projection instance used to transform given lat-lon coords to in-projection
+        Cartesian coordinates. If None the coords are assumed to already be Cartesian.
+    decode_mask: Iterator or None
+        Mask describing which grid positions should be decoded to (included in the m2g subgraph).
+        It should have the same length as the number of grid position coordinates given in `coords`.
+        The mask being set to True means that corresponding grid nodes should be included in g2m.
+        If `decode_mask=None` (default), all grid nodes are included.
```
SimonKamuk, line 35: Surely decode_mask should be an Iterable rather than an Iterator, right? Also maybe just to be completely clear, I suggested a comment about the mask specifying what is included in the output graph.

```suggestion
    decode_mask: Iterable or None
        Mask describing which grid positions should be decoded to (included in the m2g subgraph), i.e. which positions should be output.
        It should have the same length as the number of grid position coordinates given in `coords`.
        The mask being set to True means that corresponding grid nodes should be included in g2m.
        If `decode_mask=None` (default), all grid nodes are included.
```

src/weather_model_graphs/create/archetype.py
```diff
@@ -61,26 +79,35 @@ def create_graphcast_graph(
...
     max_num_levels: int
         The number of levels of longer-range connections in the mesh graph.
+    projection: cartopy.crs.CRS or None
+        Projection instance used to transform given lat-lon coords to in-projection
+        Cartesian coordinates. If None the coords are assumed to already be Cartesian.
+    decode_mask: Iterator or None
+        Mask describing which grid positions should be decoded to (included in the m2g subgraph).
+        It should have the same length as the number of grid position coordinates given in `coords`.
+        The mask being set to True means that corresponding grid nodes should be included in g2m.
+        If `decode_mask=None` (default), all grid nodes are included.
```
SimonKamuk, line 99:
```suggestion
    decode_mask: Iterable or None
        Mask describing which grid positions should be decoded to (included in the m2g subgraph), i.e. which positions should be output.
        It should have the same length as the number of grid position coordinates given in `coords`.
        The mask being set to True means that corresponding grid nodes should be included in g2m.
        If `decode_mask=None` (default), all grid nodes are included.
```

src/weather_model_graphs/create/archetype.py
```diff
@@ -119,23 +153,32 @@ def create_oskarsson_hierarchical_graph(
...
     level_refinement_factor: float
         Refinement factor between grid points and bottom level of mesh hierarchy
+    projection: cartopy.crs.CRS or None
+        Projection instance used to transform given lat-lon coords to in-projection
+        Cartesian coordinates. If None the coords are assumed to already be Cartesian.
+    decode_mask: Iterator or None
+        Mask describing which grid positions should be decoded to (included in the m2g subgraph).
+        It should have the same length as the number of grid position coordinates given in `coords`.
+        The mask being set to True means that corresponding grid nodes should be included in g2m.
+        If `decode_mask=None` (default), all grid nodes are included.
```
SimonKamuk, line 170:
```suggestion
    decode_mask: Iterable or None
        Mask describing which grid positions should be decoded to (included in the m2g subgraph), i.e. which positions should be output.
        It should have the same length as the number of grid position coordinates given in `coords`.
        The mask being set to True means that corresponding grid nodes should be included in g2m.
        If `decode_mask=None` (default), all grid nodes are included.
```

src/weather_model_graphs/create/base.py
```diff
@@ -70,27 +78,43 @@ def create_all_graph_components(
     - "containing_rectangle": For each grid node, find the rectangle with 4 mesh nodes as corners
         such that the grid node is contained within it. Connect these 4 (or less along edges)
         mesh nodes to the grid node.
+
+    `projection` should either be a cartopy.crs.CRS or None. This is the projection
+    instance used to transform given lat-lon coords to in-projection Cartesian coordinates.
+    If None the coords are assumed to already be Cartesian.
+
+    `decode_mask` should be an Iterable of booleans, masking which grid positions should be
+    decoded to (included in the m2g subgraph). It should have the same length as the number of
```
SimonKamuk, line 87:
```suggestion
    decoded to (included in the m2g subgraph), i.e. which positions should be output. It should have the same length as the number of
```

SimonKamuk (commented):
Great work! Other than remembering to update the changelog and readme, I suggested some very minor changes, for you to use if you agree with them.

src/weather_model_graphs/create/archetype.py
joeloskarsson: Yes, good catch :smile:

src/weather_model_graphs/save.py
joeloskarsson: Agreed!

joeloskarsson:
Good suggestions all around @SimonKamuk ! Was there something specific introduced by this PR that you think should be added to the readme? I was thinking that the example in the readme can continue to use default parameters, which is `decode_mask=None`.

Otherwise this is the plan for this PR now:

- [ ] Make sure v0.2.0 is released
- [ ] Make sure #32 is merged
- [ ] Add this change to changelog, under unreleased to be in v0.3.0
- [ ] Merge

SimonKamuk:
Well, it's not strictly necessary, and I think it's actually better suited for the documentation notebooks rather than the readme, but I just thought it would be nice to describe this feature with an example to make it easier for people to find. But again, it might not be necessary 😄

joeloskarsson:
I do agree though that it could be nice to have a documentation notebook with an example of using this. Will see if I can put together something small.

leifdenby:
I just had a quick skim using the diff to #32  link you provided (thanks!). In general this looks great. I will make a proper review once #32 is merged in. I hope that's ok!

joeloskarsson:
There is one small problem currently, which is one reason why https://github.com/mllam/weather-model-graphs/pull/34/commits/a6b613786be0917d05f9c3462d783f13efb51bce was introduced.

When using the decoding mask (maybe this could happen also without it, not sure?), not all mesh nodes might be connected to the grid in m2g, i.e. there are mesh nodes that are not endpoints of any edges in m2g. This is desirable. However, when you split up a graph using https://github.com/mllam/weather-model-graphs/blob/a6d43e393cd902f4c24d860366097b596852df26/src/weather_model_graphs/networkx_utils.py#L50
only nodes that are endpoint of any edge will be included. Node subsetting is only implicit through the edge sub-graphing: https://github.com/mllam/weather-model-graphs/blob/a6d43e393cd902f4c24d860366097b596852df26/src/weather_model_graphs/networkx_utils.py#L80-L88
This means that something like https://github.com/mllam/weather-model-graphs/blob/a6d43e393cd902f4c24d860366097b596852df26/tests/test_save.py#L30 will implicitly filter the mesh nodes in m2g to just the ones that are connected to the grid. This messes up the indexing of mesh nodes, and makes graphs saved this way incompatible with neural-lam.

This is why the option to return separate components was introduced, since it is unnecessary to merge everything and then split it up again (especially as splitting now breaks the node indices). Do you have any thoughts about good way to fix this @leifdenby ? Or do we want to fix it? Not sure if if has to be fixed in this PR, since in a sense it's a more general issue with the edge-attribute-based sub-graphing.

joeloskarsson:
Now all changes from #32is merged into main and here so this diff is readable.

leifdenby:
> This messes up the indexing of mesh nodes, and makes graphs saved this way incompatible with neural-lam.

Ok, that sounds like an issue. I don't fully understand why yet, but I believe you. So rather than merging to create graph that represents the whole encode-process-decode graph of nodes and edges, you introduced `return_components` in https://github.com/mllam/weather-model-graphs/commit/a6b613786be0917d05f9c3462d783f13efb51bce#diff-72daea2494478edf48e319a44c42b839b9a6175a03dd1b2fa97ca8db212a79caR184 to instead return `g2m`, `m2m` and `m2g` separately. Is that right?

> This is why the option to return separate components was introduced, since it is unnecessary to merge everything and then split it up again (especially as splitting now breaks the node indices). Do you have any thoughts about good way to fix this @leifdenby ? Or do we want to fix it? Not sure if if has to be fixed in this PR, since in a sense it's a more general issue with the edge-attribute-based sub-graphing.

The reason why I implemented the graph creation by including a merging to one big graph is so that each node to could be given a globally (as in across the whole graph) unique ID. I then assumed that this id could then be used for constructing the adjacency matrices for different components (g2m, m2m, m2g say) which are then saved to individual files. Are you saying that this of having a global index doesn't work when masking out which grid-nodes are used in the `m2g` component? Or is it because the index in the grid is actually not a global index but instead an index that only runs over grid points maybe?

I might have to create a little toy example to understand the issue properly

leifdenby:
> I might have to create a little toy example to understand the issue properly

Maybe if you have time you could add a test/notebook @joeloskarsson with an assert that checks for the indexing that you expect? I assume it is the `.pt`-file that represents the `m2g` adjacency matrix that is incorrect.

I tried creating a notebook just now, but I am not quite sure about what the correct answer would be...

joeloskarsson:
Yes, the issue does appear when you split the combined graph. But the problem is not that the indices are wrong, it is that you lose some nodes (which make the indices wrong on those remaining nodes). So your toy example could be:

1. Create a graph where not all mesh nodes are connected to the grid in m2g, i.e. there are no edges in m2g that start in these mesh nodes.
2. Split this graph as in https://github.com/mllam/weather-model-graphs/blob/a6d43e393cd902f4c24d860366097b596852df26/tests/test_save.py#L30
3. Now the m2g component will no longer contain these mesh nodes that are not connected to the grid. So e.g. g2m and m2g will have a different amount of mesh nodes.

joeloskarsson:
>  I was under the impression that torch_geometric.utils.convert.from_networkx() will use the node labels as the indicies in the adjacency matrix (and that is why I set the node labels as a globally unique index in https://github.com/mllam/weather-model-graphs/blob/main/src/weather_model_graphs/create/base.py#L176). But that I realise now that is not the case, instead the node indicies are always in the range [0,n_grid_points-1] as defined in the following line:

Yes, this is exactly what I also realized. I wish I would have had a bit more time to write out an explanation for this and I could have saved you the work :sweat: Well well, I hope you feel like you gained some insights on the way. In my view the arbitraryness of indexing when converting networkx tto pyg feels really bad and not very thought through. But I guess this is all just because networkx doesn't really have a notion of an integer node index (closer to how we think of a graph in theory, with sets, bad for practical work imo :stuck_out_tongue:).

>  I think (based on the fact that we're having this discussion) that g2m and m2g should share indices so that the same grid-point in either g2m or m2g has the same index in both.

I am not sure if that is actually necessary, but we could do that. My implementation using this in neural-lam makes the hard decision that the nodes that you want to decode to have the first indices, follows by the nodes only used in g2m. This does mean that it doesn't matter if you include the masked out node indices in your m2g graph (you'll get the same edge_index tensor in the end anyway). On a more conceptual level, one could either argue that m2g should only index into the set of nodes we decode to, or argue that m2g should index into the set of all grid nodes, but only contain edges to the unmasked ones. Not obvious to me what makes more sense.

> What I still don't quite understand is what neural-lam assumes about the grid-index values for different parts of the whole encode-process-decode graph.

Yea, there is quite a lot to this. I have made very minimal assumptions wrt node indexing in neural-lam. Given that I know that we have some different perspectives on the graph (I have this perspective of separate graphs, rather than one joined), I feel like it is something that should be explained better. Maybe I could try to write up some explanation for what neural-lam expects (in that is https://github.com/mllam/neural-lam/pull/93) when I find some time, or we could sit down and talk it through.

leifdenby:
Thanks @joeloskarsson ! I am glad we finally found out where the indexing issue is arising from :sweat_smile:

@SimonKamuk had the idea that if we want to retain the idea of building a single networkx.DiGraph object and then splitting it, what we could do is to add an option when splitting this "global" graph that indicates that the user wishes to retain in every subgraph all the nodes, but only the edges that have a specific value for an attribute. This would mean that all subgraphs that are created would share the same indexing when converting using pyg, but of course each subgraph will also contain a lot of nodes that are not connected by any edges. Would this be ok? I made the plot I've included here when I was trying to work out what you store in the pickled tensor files. And at least wrt the edge index that appears to be a globally unique integer for each node. So maybe this would be an ok approach?

One thing I'm worried about is that this would cause the `_node_features.pt` file to have the same content for every subgraph I think.
pyg_edge_index

joeloskarsson:
I think what you describe would indeed work, and solve the problem. But would have to test with neural-lam to be sure. And yes, we would have to then rework what is saved in each `_node_features.pt` so that only features for the correct subset of nodes are in there (and not all nodes). 

I think however that the easier fix is to rework https://github.com/mllam/weather-model-graphs/blob/a6d43e393cd902f4c24d860366097b596852df26/src/weather_model_graphs/networkx_utils.py#L50 
to somehow not drop the mesh nodes that are not connected to the grid. Or rather, to split on something else than edge attributes then. One option could be to introduce some sort of node attribute that one could split based on, telling which sub-graphs each node is part of. That gets slightly complicated though, as nodes can belong to multiple subgraphs (e.g. g2m and m2g). 
Then there is of course always the options to not merge all graphs just to then split them again ;) (as I implemented here)

Regarding this PR though: Are you ok with merging this with the `return_components` option, and trying to fix the splitting issue separately? Or do you see that as necessary to fix as part of this PR?

leifdenby:
> I think what you describe would indeed work, and solve the problem. But would have to test with neural-lam to be sure

Ok, great!

> Regarding this PR though: Are you ok with merging this with the `return_components` option, and trying to fix the splitting issue separately? Or do you see that as necessary to fix as part of this PR?

Good point. I think we should merge this PR with the functionality you have proposed. The functionality added here makes sense and we should keep moving. I will give a final review of the code as-is so we can get this in.

(In general wrt how the graph is stored to disk from `weather-model-graph` and loaded in `neural-lam` I am not fully sure how to split the functionality. It seems clear to me that we don't want to simply serialise `networkx.DiGraph` object to disk directly, since that would require `networkx` in `neural-lam`, on the other hand the format may evolve in neural-lam. In a way it will be nice to formalise this "graph storage format" at some point, but I will leave that pending)

joeloskarsson:
The only thing that I think is missing here is a documentation notebook showing how to use this. I have not got around to making that yet.

src/weather_model_graphs/networkx_utils.py
```diff
@@ -98,40 +98,29 @@ def split_graph_by_edge_attribute(graph, attr):
...
 
-    # check that at least one subgraph was created
-    if len(subgraphs) == 0:
-        raise ValueError(
-            f"No subgraphs were created. Check the edge attribute '{attr}'."
-        )
+def sort_nodes_in_graph(graph):
+    """
+    Creates a new networkx.DiGraph that is a copy of input, but with nodes
+    sorted according to their id
```
leifdenby, line 107:
```suggestion
    sorted according to their label value
```

I think strictly speaking nodes are identified by something called their "label" rather than "id" (e.g. https://networkx.org/documentation/stable/reference/generated/networkx.relabel.convert_node_labels_to_integers.html#networkx.relabel.convert_node_labels_to_integers). And in our case we use integers as labels

src/weather_model_graphs/networkx_utils.py
```diff
@@ -98,40 +98,29 @@ def split_graph_by_edge_attribute(graph, attr):
...
+    """
+    sorted_graph = networkx.DiGraph()
+    sorted_graph.add_nodes_from(sorted(graph.nodes(data=True)))
+    sorted_graph.add_edges_from(graph.edges(data=True))
 
-    return subgraphs
+    return sorted_graph
 
 
 def replace_node_labels_with_unique_ids(graph):
```
leifdenby, line 126: I am realising now that this is a poor name, maybe I should have called it `replace_node_labels_with_unique_integer_values`

leifdenby:
> The only thing that I think is missing here is a documentation notebook showing how to use this. I have not got around to making that yet. But maybe you could just turn the notebook you added into that?

I think the notebook you have added is already fine. Or by "this" do you not mean the decoding mask functionality or is it something else?

leifdenby (commented):
Looks great! I've just added a few suggestions

src/weather_model_graphs/create/archetype.py
```diff
@@ -18,20 +18,29 @@ def create_keisler_graph(xy_grid, grid_refinement_factor=3):
...
-        Refinement factor between grid points and mesh
+    coords: np.ndarray
+        2D array of grid point positions, either in-projection Cartesian coordinates or lat-lons
+    mesh_node_distance: float
+        Distance (in x- and y-direction) between created mesh nodes,
+        in coordinate system of coords
+    projection: cartopy.crs.CRS or None
+        Projection instance used to transform given lat-lon coords to in-projection
+        Cartesian coordinates. If None the coords are assumed to already be Cartesian.
+    decode_mask: Iterable or None
```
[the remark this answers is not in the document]
joeloskarsson: I was thinking that it gets quite annoying to maintain if we list all the arguments possible to give to `create_all_graph_components` in each archetype again. It seems to me like a good use of `**kwargs`, and then we can just make sure that everything is well-documented in `create_all_graph_components`.

I now kept the `**kwargs` but added the type hint 
https://github.com/joeloskarsson/weather-model-graphs/blob/f50a9cd85ed4850432deda1f7afa35b27c6b8f81/src/weather_model_graphs/create/base.py#L44
Let me know if you think that sounds ok.

src/weather_model_graphs/create/archetype.py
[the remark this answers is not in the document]
joeloskarsson: Note that it would not just be about adding the `decode_mask` option to each archetype, but also `coords_crs` and `graph_crs`. That feels like so much duplicate code+documentation to me, and not easy to maintain. That is why I think the `**kwargs` solution makes sense for the archetypes.

src/weather_model_graphs/networkx_utils.py
joeloskarsson: Agreed, but maybe not for this PR, as it is orthogonal.

joeloskarsson:
@leifdenby if you agree with my fixes this should now be done, except for the documentation notebook. The tests fail because of the existing test notebook `docs/decoding_mask.ipynb`. I did not want to remove that file in case you want to base the documentation on this, but it needs some cleanup and text to be a useful documentation.

PR #58, opened 2026-02-19
feat: add option to crop mesh nodes to convex hull of grid coordinates

### LINKED ISSUE #40: Add an option to crop mesh nodes to convex hull of grid coordinates
This would prevent parts of the mesh "hanging freely" above regions where there are no grid nodes. Such mesh nodes are not connected to the grid, making them useless (and it is likely bad for model performance to include them).

I have a working proof-of-concepts (maybe even something that could be used directly) of this at https://github.com/joeloskarsson/weather-model-graphs/tree/mesh_chull_filtering .

Compare before:

Image

after:

Image

(Not exactly the same grid coordinates / graph settings, but these images get the idea across 😄 )

AdMub:
## Describe your changes

This PR adds a `crop_to_convex_hull` boolean parameter to `create_single_level_2d_mesh_graph` and `create_multirange_2d_mesh_graphs`.

When set to `True`, the function computes the `scipy.spatial.ConvexHull` of the provided `xy` coordinates and uses a `scipy.spatial.Delaunay` triangulation to efficiently identify and remove any generated mesh nodes that fall outside the spatial boundary of the actual data. This prevents the generation of "dead space" nodes in the corners of bounding boxes when working with irregular or circular spatial datasets.

**Testing performed:**

The implementation was verified locally using two distinct spatial datasets to ensure the pruning logic correctly identifies the data perimeter:

- **Circle Test:** Using a circular data cluster, the standard rectangular mesh generated 400 nodes, while the convex hull cropped mesh generated 316 nodes (successfully pruning 84 empty-space nodes).

- **Irregular Shape Test:** To test robustness against more complex boundaries, an irregular "heart-shaped" dataset was used. The standard mesh generated 625 nodes, which was successfully pruned to 419 nodes.

## Visualization

The plot below demonstrates the "Before" (standard bounding box) and "After" (convex hull cropped) results for the irregular test case. Note how the mesh nodes are strictly contained within the data boundary: 

  

convex_hull_verification

## Issue Link

Closes #40

## Type of change

- [x] I have given the PR a name that clearly describes the change, written in imperative form (context).

## Checklist for reviewers

- [ ] the code is readable

- [ ] the code is well tested

- [ ] I have added a line to the CHANGELOG describing this change, in a section reflecting type of change (add section where missing)

## Checklist for assignee

- [ ] the tests pass

- [ ] author has added an entry to the changelog (and designated the change as `added`, `changed` or `fixed`)


joeloskarsson (changes requested):
This option needs to be made available for all types of graphs, and when making them using the different archetypes. You can have a look at my prototype implementation on https://github.com/joeloskarsson/weather-model-graphs/tree/mesh_chull_filtering for some inspiration.

leifdenby:
I'm adding this to milestone `v0.4.0` since the issue it is resolving (#40) is already on the roadmap for that release. We can always revisit whether to postpone this for a later release

leifdenby (approved):
Looking good! Some comments for suggestions

prajwal-tech07:
Hey @AdMub , just a heads-up that PR #81 has now been merged into main! 

Since you were waiting on that merge before syncing your branch and pushing the updated version , that's now unblocked! It would be great if you could rebase off the updated main, move `crop_to_convex_hull` into `mesh_layout_kwargs` as @leifdenby suggested, add the docstring to `create_single_level_2d_mesh_graph`, and push the clean version to get CI green.

This PR is on the v0.4.0 milestone and it's a blocker for the release, so getting it moving again would really help. If you're currently busy or unable to continue, no worries at all , just let us know and I can take this forward so we don't hold up the v0.4.0 release. Thanks for all the work so far!

AdMub:
Hi @prajwal-tech07, thanks for the heads-up! I am on it.
I will rebase against the updated `main`, resolve the merge conflicts in `mesh.py`, and shift `crop_to_convex_hull` into `mesh_layout_kwargs` as @leifdenby suggested. I'll also make sure the docstrings are updated.
I'll push the clean version shortly so we can get the CI green and unblock the v0.4.0 release! Thanks for keeping an eye on this.

leifdenby:
@AdMub thanks for working on this! There is just one merge conflict to resolve, do you think you would have time to look at this? Once that is done I will get the tests to run through and give this a review :)

Thanks again!

AdMub:
Hi @leifdenby, apologies for that! `main` must have moved forward a bit. I will pull the absolute latest, resolve that last conflict in `mesh.py`, and push it up right now!

convex_hull_verification.png
joeloskarsson, line 1: Nice, but should not be merged :)

joeloskarsson (changes requested):
Happy to review this again when up to date with main. Please re-request a review from me.

prajwal-tech07:
Hi @AdMub! I took a look at why the CI is failing - there are two separate issues:

1. The tests job fails at collection (this is the main blocker). There are some debug/scratch files committed at the repo root that shouldn't be in the PR:

test_issue_44.py - imports weather_model_graphs.create.mesh.kinds.flat, which doesn't exist on this branch → ModuleNotFoundError, so pytest aborts collection and the whole suite fails before any test runs.
test_issue_40.py, visualize_fix_40.py, convex_hull_verification.png - these look like local debugging artifacts too.
Could you remove those four from the repo root? The proper test already lives in tests/test_graph_creation.py, which is the right place.

2. After removing those, the actual feature test still fails. test_convex_hull_cropping errors with:


TypeError: create_keisler_graph() got an unexpected keyword argument `mesh_layout_kwargs`
It looks like crop_to_convex_hull / mesh_layout_kwargs is added to the lower-level mesh functions but isn't threaded through the archetype API (create_keisler_graph → create_all_graph_components), so the feature can't be reached the way the test calls it. That plumbing needs to be connected.

One bigger-picture thing: the cropping code is added as a new src/.../mesh/mesh.py, but main reorganized mesh creation in #81 - so it'd be worth integrating into the current structure rather than a parallel module. Happy to help if useful!

AdMub:
Hi @prajwal-tech07, you hit the nail on the head. I didn't realize `mesh.py` was a deprecated module that got resurrected during my git rebase, that completely explains why my tests were asserting `100 < 100` (the Delaunay pruning logic wasn't even executing!).

I just pushed a commit to force-delete the `mesh.py` ghost file, plumb the `mesh_layout_kwargs` correctly through the archetypes, and fix all the CI formatting. As expected, the test still fails because the logic is removed.

I would love your help integrating the `ConvexHull` pruning into the new #81 structure. Could you point me to the exact file/function where the mesh nodes are generated now so I can inject the spatial pruning logic there? Thanks so much

prajwal-tech07:
Glad that explained it! Here's where mesh nodes are generated after #81, and an important caveat I found while testing it.

**Where:** it's all in create_all_graph_components in src/weather_model_graphs/create/base.py, in the **"Step 1: Coordinate creation"** block (around lines 258–290). Grid coords are in the local xy, and the mesh nodes are built into G_mesh_coords by create_single_level_2d_mesh_primitive / create_multirange_2d_mesh_primitives (both in mesh/coords.py). The clean place to inject the pruning is r**ight after G_mesh_coords is created, before the Step 2 connectivity block**, since you have both xy and the mesh positions there. Something like:


```
if mesh_layout_kwargs.get("crop_to_convex_hull", False):
    hull = scipy.spatial.Delaunay(xy)
    def _crop(g):
        outside = [n for n, pos in g.nodes(data="pos") if hull.find_simplex(pos) < 0]
        g.remove_nodes_from(outside)
        return g
    if isinstance(G_mesh_coords, list):
        G_mesh_coords = [_crop(g) for g in G_mesh_coords]
    else:
        G_mesh_coords = _crop(G_mesh_coords)
```
**Important caveat I hit while testing this:** I tried this across all three archetypes. It works for **keisler (flat)** and **hierarchical,** but it breaks graphcast **(flat_multiscale**) with ValueError: cannot reshape array of size ... into shape (...). The reason is create_flat_multiscale_from_coordinates reshapes the node array assuming a full rectangular grid, which no longer holds once nodes are cropped.

So I'd suggest: for now, either scope the feature to flat/hierarchical and raise a clear error if crop_to_convex_hull is used with flat_multiscale, or adapt that connectivity function to not assume a regular grid (bigger change). Either way, I'd extend the test to cover all three archetypes, not just keisler - otherwise this break stays hidden. Happy to help with whichever direction you pick!

AdMub:
Hi @leifdenby, @prajwal-tech07 and @joeloskarsson, the PR is now completely green!

The Convex Hull pruning is successfully integrated into `base.py` and correctly skips the unsupported `flat_multiscale` architecture. The tests have been expanded to cover this and all formatting checks have passed.

I have re-requested a review—let me know if there's anything else needed before we merge this into `v0.4.0`!

src/weather_model_graphs/create/base.py
```diff
@@ -289,6 +289,31 @@ def create_all_graph_components(
             "Currently only 'rectilinear' is implemented."
         )
 
+    # --- CONVEX HULL PRUNING INJECTION ---
+    if mesh_layout_kwargs and mesh_layout_kwargs.get("crop_to_convex_hull", False):
+        if m2m_connectivity == "flat_multiscale":
+            raise ValueError(
+                "crop_to_convex_hull is currently not supported for 'flat_multiscale' "
+                "(GraphCast) architectures because they assume a regular rectangular grid."
+            )
```
joeloskarsson, line 306: How tricky would it be to support multiscale graphs here? I think it would be nice, but I am probably missing something about what makes it difficult. In particular, why can we not just crop G_mesh_coords in a similar way for flat_multiscale graphs?

prajwal-tech07:
Hi @joeloskarsson - I looked into this while reviewing, so I can explain the difference. The flat path connects nodes purely by position, so cropping the nodes first just works. flat_multiscale has an extra step that merges the levels by index, not position: in `create_flat_multiscale_from_coordinates` it reshapes the node list into the full rectangular grid and strides over it (`reshape((nx, ny, 2))[offset::factor]`) to align the coarser levels onto the finest. That assumes every grid position is present, so once nodes are cropped the reshape fails.

Supporting multiscale would mean changing that merging to be position-based instead of index-based - doable, but a bigger change that's probably best as a separate follow-up rather than part of this PR.

AdMub:
Hi @joeloskarsson and @prajwal-tech07,
Prajwal is exactly right, thank you for breaking down that `reshape`constraint! Because `create_flat_multiscale_from_coordinates` relies on strict array striding to map coarser levels, cropping the nodes beforehand breaks the `(nx, ny, 2)` dimensional assumption.

I completely agree with you, Joel, that it would be highly valuable to support multiscale graphs. Shifting that merging logic from index-based to spatial/position-based is definitely doable. However, since it requires altering the core connectivity assumptions of `flat_multiscale`, I agree with Prajwal that it's best handled as a dedicated follow-up PR to avoid scope creep on this one.

leifdenby:
> I completely agree with you, Joel, that it would be highly valuable to support multiscale graphs. Shifting that merging logic from index-based to spatial/position-based is definitely doable. However, since it requires altering the core connectivity assumptions of `flat_multiscale`, I agree with Prajwal that it's best handled as a dedicated follow-up PR to avoid scope creep on this one.

@AdMub might it be useful to add an inline comment for where this change would need to be made down the line to generalise the implementation?

AdMub:
Hi @leifdenby, that is a great suggestion.
I have just pushed a commit adding a detailed `TODO` comment right above the `ValueError`in `base.py`. It points specifically to `create_flat_multiscale_from_coordinates` and documents the need to refactor the index-based array reshaping `(nx, ny, 2)` into a position-based spatial merge.

That should leave a very clear breadcrumb for the follow-up PR. Let me know if that looks good to go!

leifdenby (commented):
> Hi @leifdenby, that is a great suggestion. I have just pushed a commit adding a detailed `TODO` comment right above the `ValueError`in `base.py`. It points specifically to `create_flat_multiscale_from_coordinates` and documents the need to refactor the index-based array reshaping `(nx, ny, 2)` into a position-based spatial merge.
> 
> That should leave a very clear breadcrumb for the follow-up PR. Let me know if that looks good to go!

great! Thank you 🌟 Do you fancy typing up an issue too describing the follow-up that should be done after we've merged this PR? It is good you implemented the exception to make it clear that "cropping to convex hull" doesn't work for flat-multiscale graphs.

I think something we're missing here is a jupyter notebook that demonstrates this new functionality. What it makes sense to introduce a subsection to the existing notebooks? Or do you think it would be good with a new separate notebook? I think if we don't add a notebook here this feature might simply go unnoticed which would be a shame.

docs/creating_the_graph.ipynb
leifdenby, line 1: Ok, this is clearer to me now you have added the text in the notebook :) Could you also plot the grid nodes in your figure? Also, what do you think about changing the argument from `crop_to_grid_nodes_convex_hull`? I also think it would be better to put this in a separate notebook since the current `creating_the_graph.ipynb` notebook is quite long already.

leifdenby (changes requested):
One last suggestion :) Will you also add an entry to the CHANGELOG explaining your contribution? Thanks!

docs/creating_the_graph.ipynb
AdMub: Hi @leifdenby, I am happy to make those documentation changes!
Just to clarify on the argument name: did you mean changing it to `crop_to_grid_nodes_convex_hull`, or were you suggesting we change it from `crop_to_convex_hull` to something else?

Let me know your preferred name and I'll do a global rename, split the notebook, and update the changelog in my next push!"

leifdenby: Sorry, I should've reread what I wrote before I pressed <Comment> :)

I meant change the name of the argument from `crop_to_convex_hull` to `crop_to_grid_nodes_convex_hull` since the latter is explicit about the convex hull of *what* is being used for cropping. What do you think?

Thanks for being happy with making these changes!

AdMub:
"Hi @leifdenby, that makes perfect sense! `crop_to_grid_nodes_convex_hull` is definitely more explicit and leaves no room for ambiguity.
I have done a global find-and-replace to update the argument name across the codebase, tests, and the new notebook. I also synced with `main` and resolved the merge conflict in the `CHANGELOG` to make sure the new feature is documented correctly.
Pushed and running CI now. I think this should be the final piece of the puzzle for `v0.4.0`!

src/weather_model_graphs/create/archetype.py
```diff
@@ -13,6 +13,7 @@ def create_keisler_graph(
     graph_crs: pyproj.crs.CRS | None = None,
     decode_mask: Iterable[bool] | None = None,
     return_components: bool = False,
+    mesh_layout_kwargs: dict | None = None,
```
leifdenby: I don't think we should expose the whole `mesh_layout_kwargs` here, doing that exposes a lot of complexity the most people will not need and we will have to greatly increase the content of the docstring. Instead if you would like to expose `crop_to_grid_nodes_convex_hull` through the archetype functions I would expose just that argument and insert it into the `mesh_layout_kwargs`, just like what we do for `mesh_node_spacing`

docs/mesh_cropping.ipynb
leifdenby, line 1: the `pruned mesh` nodes appear identical to the mesh nodes before pruning, did you inadvertently introduce a bug in your implementation here? :)

leifdenby (changes requested):
Hmm... I think something has gone wrong in the notebook or the implementation (see my comment).

Also, I think we should not expose the whole `mesh_layout_kwargs` in the archtype functions, let me know what you think :)

AdMub:
Hi @leifdenby, you were 100% right on both counts!
Because I was passing `mesh_layout_kwargs` as a dictionary from the outside, the internal archetype logic was overwriting it, which completely bypassed the cropping in the notebook plot. Great catch!
I have refactored `archetype.py` so that `crop_to_grid_nodes_convex_hull` is exposed explicitly as a boolean on the archetype signatures. The functions now safely pack it into `mesh_layout_kwargs` internally.
I've also updated the notebook to use the explicit boolean argument, and the pruned mesh visualization is now rendering perfectly. Pushed and running CI!

src/weather_model_graphs/create/archetype.py
```diff
@@ -56,12 +57,17 @@ def create_keisler_graph(
     networkx.DiGraph or dict[networkx.DiGraph]
         The graph or graph components.
     """
+    kwargs = dict(
```
leifdenby: can you revert the change here so that rather than creating a dict for the kwargs and calling it `kwargs` here we create the `dict` inline as before?

src/weather_model_graphs/create/archetype.py
```diff
@@ -128,16 +135,19 @@ def create_graphcast_graph(
...
```
leifdenby: same comment as above, create the dict inline where passing arguments to `create_all_graph_components` rather than creating the intermediate `kwargs` variable

src/weather_model_graphs/create/archetype.py
```diff
@@ -209,16 +220,19 @@ def create_oskarsson_hierarchical_graph(
...
```
leifdenby: as above

src/weather_model_graphs/create/base.py
leifdenby: @AdMub would you be able to reply this comment?

CHANGELOG.md
```diff
@@ -29,9 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (`save.base`, `save.neural_lam.torch_tensors`, `save.neural_lam.deprecated`)
   with all existing entrypoints re-exported unchanged.
   [\#123](https://github.com/mllam/weather-model-graphs/pull/123), @prajwal-tech07
-- Add support for writing benchmarking results to json,
```
leifdenby, line 32: You've deleted a changelog entry here, and you need to add a changelog entry for your own work too :)

leifdenby (changes requested):
Nearly there!

leifdenby:
I've moved this feature to the next release milestone (`v0.5.0`) so that we can get `v0.4.0` out

AdMub:
Hi @leifdenby,
Good call on inlining those dictionaries! I have reverted `archetype.py` so the `mesh_layout_kwargs` are explicitly constructed inline within the `create_all_graph_components` calls, just like before.
I also completely reset the `CHANGELOG.md` to match `main` and carefully re-inserted my single line to ensure nothing else got deleted this time. All pushed and ready for `v0.5.0`!

PR #60, opened 2026-02-20
fix: standardize 'level' and 'levels' attributes across all edge directions

### LINKED ISSUE #45: graph edges `level` attribute is inconsistent across different `direction` types
The level attribute within a mesh level, e.g. (`direction == same`), is called "level", opposed to when the edges connect different mesh levels. In those cases it is called "levels".

E.g.

https://github.com/mllam/weather-model-graphs/blob/3d9d1ef670eae5e2a2443202576c20514e48f746/src/weather_model_graphs/create/mesh/kinds/hierarchical.py#L108

vs.

https://github.com/mllam/weather-model-graphs/blob/3d9d1ef670eae5e2a2443202576c20514e48f746/src/weather_model_graphs/create/mesh/kinds/hierarchical.py#L68

This causes issues, e.g. when using https://github.com/pyg-team/pytorch_geometric/blob/afffee9f089ac0fe1eb8c1600c6e0b2e07ca6f80/torch_geometric/utils/convert.py#L260-L262

AdMub:
## Description
This PR resolves #45 by standardizing the `level` (integer) and `levels` (string) attributes across all edge direction types (`same`, `up`, `down`) in the hierarchical multiscale mesh graph.

**Previously:**
* `"same"` edges lacked the `levels` string attribute.
* `"up"` and `"down"` edges lacked the `level` integer attribute entirely. 

**The Fix:**
* `"same"` edges are now explicitly assigned `levels = f"{i}>{i}"`.
* `"down"` and `"up"` edges are now assigned `level = max(from_level, to_level)`. Assigning inter-level edges to the coarser level ensures that graph traversal logic expecting uniform integer levels will process the hierarchy correctly, and prevents `level == 0` weather data from attaching incorrectly.

## Testing
Verified locally by generating a hierarchical multiscale mesh graph (`max_num_levels=3`) and asserting the existence of both attributes across all direction types.

**Output:**
```text
Direction: same  | level: 0 | levels: 0>0
Direction: up    | level: 1 | levels: 0>1
Direction: down  | level: 1 | levels: 1>0

### DIFF
--- a/src/weather_model_graphs/create/mesh/kinds/hierarchical.py
+++ b/src/weather_model_graphs/create/mesh/kinds/hierarchical.py
@@ -66,6 +66,7 @@ def create_hierarchical_multiscale_mesh_graph(
         for u, v in G.edges:
             G.edges[u, v]["direction"] = "same"
             G.edges[u, v]["level"] = i
+            G.edges[u, v]["levels"] = f"{i}>{i}"  # <--- NEW ADDITION
 
     # Create inter-level mesh edges
     up_graphs = []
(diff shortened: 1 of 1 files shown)


AdMub:
Thanks for the update @observingClouds! That makes total sense to stick with the planned implementation for #45. I'll keep an eye out for that PR when it drops. Cheers!

PR #91, opened 2026-03-04
Add mesh_layout='prebuilt' for user-provided mesh graphs

prajwal-tech07:
## Describe your changes

Add support for `mesh_layout="prebuilt"` in `create_all_graph_components`, enabling users to supply their own mesh graph (e.g. ICON icosahedral, MPAS Voronoi, custom observation networks) instead of relying on library-generated meshes.

**Two modes are supported:**

- **Nodes+edges mode:** User provides a complete `nx.DiGraph` with validated node attributes (`pos`, `type`) and edge attributes (`len`, `vdiff`). The graph is used as-is after validation.
- **Nodes-only mode:** User provides an `nx.Graph` with only nodes (having `pos` and `type` attributes). The library automatically builds connectivity edges using Delaunay triangulation (with KDTree fallback for collinear/degenerate cases).

**Works with all three `m2m_connectivity` types:** `flat`, `flat_multiscale`, and `hierarchical`.

**Changes across 4 files:**

| File | Change |
|---|---|
| `src/.../mesh/kinds/prebuilt.py` (new, 453 lines) | Core module: validation functions, Delaunay edge builder, flat/multiscale/hierarchical constructors |
| `src/.../create/base.py` (+197/-38) | Dispatch logic in `create_all_graph_components` for all 3 `m2m_connectivity` branches |
| `src/.../mesh/__init__.py` (+9) | Public re-exports of prebuilt functions |
| `tests/test_prebuilt_mesh.py` (new, 2869 lines) | **202 tests** across 28 test classes covering validation, Delaunay, flat/multiscale/hierarchical, integration, edge cases, numerical correctness, graph invariants, and stress tests |

**Motivation:** This is a key extensibility feature that decouples mesh topology from the graph construction pipeline, allowing researchers to use arbitrary mesh structures without modifying library internals.

**Dependencies:** No new dependencies — uses existing `networkx`, `numpy`, `scipy.spatial` (Delaunay, KDTree).

## Issue Link

Closes #79

## Type of change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation (Addition or improvements to documentation)

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch - if not update your fork with the changes from the target branch (use `pull` with `--rebase` option if possible).
- [x] I have performed a self-review of my code
[...]

PR #81, opened 2026-03-02
Add mesh_layout argument with rectilinear two-step coordinate/connectivity architecture

prajwal-tech07:
### Add `mesh_layout` argument with rectilinear two-step coordinate/connectivity architecture
Implements the `mesh_layout` argument for mesh graph creation, starting with `rectilinear` as the first supported layout. This uses a **two-step architecture** that separates:

1. **Coordinate creation** (`nx.Graph`) — places mesh nodes on a grid with `cardinal` and `diagonal` edge annotations
2. **Connectivity creation** (`nx.DiGraph`) — filters edges by pattern (`"4-star"` or `"8-star"`) and computes edge features

### Changes by file:

**`base.py`**
- Added `mesh_layout` (default: `"rectilinear"`) and `mesh_layout_kwargs` parameters to `create_single_level_2d_mesh_graph` and `create_multi_level_2d_mesh_graph`
- Shared parameter names: `refinement_factor` and `max_num_refinement_levels`
- `flat_multiscale` dispatches with simple `pattern` kwarg (no sub-dicts)
- `hierarchical` dispatches with `intra_level`/`inter_level` sub-dicts
- Backward compatibility: old parameter names (`mesh_node_distance`, `level_refinement_factor`, `max_num_levels`) still work with deprecation warnings
- Validates `grid_spacing` presence when `mesh_layout` is specified

**`flat.py`**
- `create_flat_multiscale_from_coordinates()` now accepts a simple `pattern="8-star"` argument instead of sub-dicts, matching Leif's API specification

**`archetype.py`**
- Updated `create_graphcast_graph` and `create_oskarsson_hierarchical_graph` to use new parameter names (`refinement_factor`, `max_num_refinement_levels`)

**`tests/test_mesh_layout.py`**
- 81 comprehensive tests covering coordinate creation, connectivity creation, API integration, backward compatibility, error handling, archetype equivalence, and 36 edge case tests

### API table (per specification in #78):

| `m2m_connectivity` | `mesh_layout` | `mesh_layout_kwargs` |
|---|---|---|
| `flat` | `rectilinear` | `pattern="8-star"` |
| `flat_multiscale` | `rectilinear` | `pattern="8-star"`, `refinement_factor=3`, `max_num_refinement_levels=None` |
| `hierarchical` | `rectilinear` | `intra_level={"pattern": "8-star"}`, `inter_level={"pattern": "knn", "k": 3}`, `refinement_factor=3`, `max_num_refinement_levels=None` |

### Motivation and context:
[...]

PR #92, opened 2026-03-05
Add mesh_layout='triangular' for regular triangular mesh generation

### LINKED ISSUE #80: [Feat] Add mesh_layout="triangular" for regular triangular mesh generation
## Summary
Add support for mesh_layout="triangular" in `create_all_graph_components()`, creating regular triangular meshes using `networkx.triangular_lattice_graph`. This mirrors the existing rectilinear mesh functions with zero new dependencies.

## Motivation
A regular triangular mesh has several advantages over rectilinear:

- 6-connectivity (vs 8 for rectilinear with diagonals) — more isotropic message passing
- Better approximation of circular receptive fields — equilateral triangles tile more uniformly than squares + diagonals
- Natural stepping stone toward icosahedral meshes on a sphere
- Relevant for LAM models — some regional NWP grids use triangular meshes

As discussed in #71, mesh_layout="triangular" means "uniform equilateral triangles in the coordinates of the provided CRS" — a regular planar grid, not Delaunay triangulation of irregular points.

## Proposed API
```
# Flat triangular mesh
graph = create_all_graph_components(
    coords=xy,
    m2m_connectivity="flat",
    m2m_connectivity_kwargs=dict(mesh_node_distance=3.0),
    mesh_layout="triangular",
    ...
)

# Hierarchical triangular mesh
graph = create_all_graph_components(
    coords=xy,
    m2m_connectivity="hierarchical",
    m2m_connectivity_kwargs=dict(mesh_node_distance=3.0, max_num_levels=3),
    mesh_layout="triangular",
    ...
)
```
## Implementation: Mirrored Function Pairs

The triangular layout mirrors the existing rectilinear functions exactly:

| Rectilinear (existing) | Triangular (new) |
|---|---|
| `create_single_level_2d_mesh_graph` uses `networkx.grid_2d_graph` (8-connectivity) | `create_single_level_2d_triangular_mesh_graph` uses `networkx.triangular_lattice_graph` (6-connectivity) |
| `create_flat_singlescale_mesh_graph(xy, mesh_node_distance)` | `create_flat_singlescale_triangular_mesh_graph(xy, mesh_node_distance)` |
| `create_flat_multiscale_mesh_graph(xy, mesh_node_distance, ...)` | `create_flat_multiscale_triangular_mesh_graph(xy, mesh_node_distance, ...)` |

Core function: create_single_level_2d_triangular_mesh_graph
```
def create_single_level_2d_triangular_mesh_graph(xy, nx, ny):
    """
    Create directed graph with triangular lattice topology.
    Uses networkx.triangular_lattice_graph which produces offset rows
    with equilateral triangle side length 1 and row spacing sqrt(3)/2.
    """
    xm, xM = np.amin(xy[:, 0]), np.amax(xy[:, 0])
    ym, yM = np.amin(xy[:, 1]), np.amax(xy[:, 1])

    dx = (xM - xm) / nx
    dy = (yM - ym) / ny

    g = networkx.triangular_lattice_graph(ny, nx, with_positions=True)

    # Scale and offset positions to cover the domain
    for node in g.nodes:
        pos = np.array(g.nodes[node]["pos"])
        # Scale from unit triangles to domain
        g.nodes[node]["pos"] = np.array([
            xm + dx/2 + pos[0] * dx,
            ym + dy/2 + pos[1] * dy
        ])
        g.nodes[node]["type"] = "mesh"

    # Convert to DiGraph with bidirectional edges
    dg = networkx.DiGraph(g)
    for u, v in g.edges():
        d = np.sqrt(np.sum((dg.nodes[u]["pos"] - dg.nodes[v]["pos"]) ** 2))
        dg.edges[u, v]["len"] = d
        dg.edges[u, v]["vdiff"] = dg.nodes[u]["pos"] - dg.nodes[v]["pos"]
        dg.add_edge(v, u)
        dg.edges[v, u]["len"] = d
        dg.edges[v, u]["vdiff"] = dg.nodes[v]["pos"] - dg.nodes[u]["pos"]

    return dg
```

`mesh_node_distance` interface
Same as rectilinear. For triangular, `mesh_node_distance` controls the side length:

```
nx = int(range_x / mesh_node_distance)
ny = int(range_y / (mesh_node_distance * np.sqrt(3) / 2))
```

### CRS Warning

If `graph_crs.is_geographic == True` and mesh_layout="triangular", emit a warning:

```
if graph_crs is not None and graph_crs.is_geographic and mesh_layout == "triangular":
    warnings.warn(
        "mesh_layout='triangular' produces non-uniform physical spacing in "
        "geographic coordinates. Consider mesh_layout='icosahedral' for "
        "uniform coverage on a sphere."
    )

```

### Depends On

- mesh_layout argument introduction issue

### Related

- #71 — Parent discussion
- #77 — mesh_layout=icosahedral (for geographic/spherical use cases)
- PR `neural-lam-dev#13` — Reference implementation for triangular graphs

prajwal-tech07:
## Describe your changes

Adds support for `mesh_layout="triangular"` in `create_all_graph_components`, using `networkx.triangular_lattice_graph` to produce an equilateral-triangle lattice with 6-connectivity (vs 8 for rectilinear). This provides more isotropic message passing for weather model graph neural networks.

**New module: `triangular.py`** (428 lines)
- `create_single_level_2d_triangular_mesh_primitive` — coordinate creation using `triangular_lattice_graph`, scales/offsets positions to cover the coordinate domain
- `create_multirange_2d_triangular_mesh_primitives` — multi-level triangular primitives at progressively coarser resolutions
- `create_single_level_2d_triangular_mesh_graph` — convenience wrapper (primitive → directed)
- `create_flat_singlescale_triangular_mesh_graph` — flat single-scale wrapper
- `create_flat_multiscale_from_triangular_coordinates` — **position-based (KDTree) node merging** for flat multiscale (replaces the index-based reshaping approach from rectilinear which is incompatible with triangular lattice node labeling)
- `create_flat_multiscale_triangular_mesh_graph` — flat multiscale wrapper
- `create_hierarchical_triangular_mesh_graph` — hierarchical wrapper

**Dispatch in `base.py`:**
- All 3 `m2m_connectivity` modes (`flat`, `hierarchical`, `flat_multiscale`) now support `mesh_layout="triangular"`
- Default connectivity pattern is `"4-star"` for triangular (all edges are cardinal; `"4-star"` == `"8-star"`)
- Added CRS geographic coordinate warning

**Tests: 97 tests** in `test_triangular_mesh.py` covering:
- Primitive creation (node count, positions, adjacency_type, type attrs, edge cases)
- Directed graph (bidirectional edges, len/vdiff attrs, 6-connectivity, degree checks)
- Multirange primitives (levels, refinement factors, domain coverage)
- Flat single-scale, flat multiscale, and hierarchical wrappers
- Full integration through `create_all_graph_components` for all 3 modes
- Error handling (missing params, spacing too large, unsupported layouts)
- Numerical correctness (len symmetry, vdiff reciprocity, no NaN/Inf, scaling)
- Rectilinear regression tests (all 3 modes still work)

This is part of the flexible graph construction effort for Issue #80.

No new dependencies — uses existing `networkx`, `numpy`, `scipy.spatial`.

**NOTE**: This PR should be adapted once `weather-model-graphs==0.4.0` has been released (where the writing to the tensor format will be included) so that released version of the package is used rather than the branch currently being pointed to.

## Issue Link

closes https://github.com/mllam/weather-model-graphs/issues/80

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


leifdenby (changes requested):
This looks great @prajwal-tech07!

Could you a notebook that demonstrates changing the mesh layout? You can just use the Keisler archetype for this, and contrast creating a graph with the default mesh layout (which would be rectilinear), one where you change the mesh-spacing, and one example where you switch to a triangular mesh.

docs/mesh_layout.ipynb
leifdenby, line 1: instead of "and can improve message-passing in graph neural network weather models" I would say "and is expected to improve message-passing in graph neural network weather models"

leifdenby: for the side-by-side comparison please set all axis to have the same limits in each subplot

leifdenby: the axes should also have aspect ratio of `1.0`

leifdenby (changes requested):
looking good! A few suggestions for the notebook

leifdenby (commented):
I think we need to rethink the module layout here. I don't think it quite make sense to consider "triangular" as being related to the connectivity, instead, the "triangular" mesh layout we are discussing here is about where the mesh coordinates are placed, rather than about the connectivity.

Isn't it the case that everything in the current `wmg.create.mesh.coords` is actually for creating mesh coordinates with the rectilinear layout? So maybe we should instead put this functionality into `wmg.create.mesh.layout.rectilinear` and then with this PR introduce `wmg.create.mesh.layout.triangular`?

I don't think we should need [`create_hierarchical_triangular_mesh_graph()`](https://github.com/prajwal-tech07/weather-model-graphs/blob/feat/triangular-mesh-layout/src/weather_model_graphs/create/mesh/connectivity/triangular.py#L366), as in a specialised method for creating a hierarchical graph from the coordinate primitives, instead it would be better if we could make use of [`create_hierarchical_from_coordinates`](https://github.com/prajwal-tech07/weather-model-graphs/blob/feat/triangular-mesh-layout/src/weather_model_graphs/create/mesh/connectivity/hierarchical.py)

What do you think?

prajwal-tech07:
That's a valid point! You're right that `wmg.create.mesh.coords` is entirely rectilinear-specific - all the coordinate creation functions there (`create_single_level_2d_mesh_primitive`, `create_multirange_2d_mesh_primitives`, etc.) are only about the rectilinear layout.

Reorganising to wmg.create.mesh.layout.rectilinear and `wmg.create.mesh.layout.triangular` makes sense -"layout" captures both where the nodes are placed and how they connect, which is exactly what each of these modules does.

Regarding `create_hierarchical_triangular_mesh_graph` - agreed, it doesn't need to be standalone. It currently just wraps `create_multirange_2d_triangular_mesh_primitives` + `create_hierarchical_from_coordinates`, which is the same pattern the rectilinear path already uses. So it can be dropped and `create_hierarchical_from_coordinates` can be reused directly with the triangular primitives.

Should I go ahead with this restructuring as part of this PR, or would you prefer it as a follow-up?

leifdenby:
> That's a valid point! You're right that `wmg.create.mesh.coords` is entirely rectilinear-specific - all the coordinate creation functions there (`create_single_level_2d_mesh_primitive`, `create_multirange_2d_mesh_primitives`, etc.) are only about the rectilinear layout.
> 
> Reorganising to wmg.create.mesh.layout.rectilinear and `wmg.create.mesh.layout.triangular` makes sense -"layout" captures both where the nodes are placed and how they connect, which is exactly what each of these modules does.
> 
> Regarding `create_hierarchical_triangular_mesh_graph` - agreed, it doesn't need to be standalone. It currently just wraps `create_multirange_2d_triangular_mesh_primitives` + `create_hierarchical_from_coordinates`, which is the same pattern the rectilinear path already uses. So it can be dropped and `create_hierarchical_from_coordinates` can be reused directly with the triangular primitives.
> 
> Should I go ahead with this restructuring as part of this PR, or would you prefer it as a follow-up?

ah, great! Ok, we are on the same page then and I wasn't just thinking down a strange route.

Yes, if you are happy with the changes too then I would say go ahead and make them, then we will be ready for our second kind of coordinate layout option in `weather-model-graphs` :D

prajwal-tech07:
Hey @leifdenby, thanks for the review! I've reorganized the module structure as requested:

- Created `wmg.create.mesh.layout` subpackage with `rectilinear` and `triangular` modules
- Moved coordinate-creation functions from `coords.py` → `layout/rectilinear.py` and from `connectivity/triangular.py` → `layout/triangular.py`
- Dropped triangular-specific convenience wrappers ,the generic connectivity functions handle triangular primitives directly
- `connectivity/triangular.py` now contains only the position-based merging logic
- Retained backward-compatible re-exports
- Updated all imports and rewrote tests to use the two-step API

All tests pass.

docs/mesh_layout.ipynb
leifdenby, line 1: In the intro here we should say that `mesh_layout` is supported from `v0.5.0` rather than `v0.4.0` :)

src/weather_model_graphs/create/mesh/connectivity/triangular.py
```diff
@@ -0,0 +1,103 @@
...
+    position coincides (within floating-point tolerance) with an existing finer
+    level node is merged with it, so that multi-resolution edges share the
+    same node identity.
+
+    Parameters
+    ----------
+    G_coords_list : list[networkx.Graph]
+        One undirected triangular mesh primitive per level.
+    pattern : str
+        Connectivity pattern: ``"4-star"`` or ``"8-star"`` (default ``"4-star"``).
```
leifdenby: Does `4-star` and `8-star` even make sense here? For a triangular mesh isn't it only `6-star` that makes sense?

src/weather_model_graphs/create/mesh/layout/__init__.py
```diff
@@ -0,0 +1,16 @@
+"""
+Mesh layout modules.
+
+Each layout module defines how mesh node coordinates are placed in space
+(coordinate creation step).  The resulting undirected primitive graphs are
+then consumed by the connectivity modules to produce directed mesh graphs.
+
+Available layouts:
+
+- ``rectilinear``: Uniform rectangular grid (``grid_2d_graph``).
```
leifdenby: what is `grid_2d_graph` and `triangular_lattice_graph` refer to here?

src/weather_model_graphs/create/mesh/layout/triangular.py
```diff
@@ -0,0 +1,218 @@
...
+   edge attributes.
+"""
+
+from typing import List
+
+import networkx
+import numpy as np
+from loguru import logger
+
+from ..connectivity.general import create_directed_mesh_graph
```
leifdenby: hmm... it is a bit odd that we're importing something from `.mesh.connectivity` here. Maybe `created_directed_mesh_graph` should actually move to `.mesh.layout.base` then? What do you think? We are arguing, I think, that connectivity is applied after constructing the layout, but here we are calling something from the "connectivity" submodule from the "layout" submodule, which implies some form of "connectivity" being applied before constructing the "layout" is complete. That doesn't seem quite logical to me, so maybe we should consider creating the directed graph as part of creating the layout? Or constructing the digraphs should be moved out of `.mesh.layout` (as in `create_single_level_2d_triangular_mesh_graph` and equivalent for the rectilinear layout should move to `.mesh.connectivity` instead of being in `.mesh.layout`)

src/weather_model_graphs/create/mesh/__init__.py
```diff
@@ -1,6 +1,13 @@
-from .coords import (
-    create_directed_mesh_graph,
+from .connectivity.general import create_directed_mesh_graph
+from .connectivity.triangular import create_flat_multiscale_from_triangular_coordinates
+from .layout.rectilinear import (
+    create_multirange_2d_mesh_graphs,
```
leifdenby, line 4: this is in line with my comment above about where the `DiGraph` objects are created. It seems odd to me that we are importing `create_single_level_2d_mesh_*graph*` and `create_single_level_2d_mesh_*primitive*` from the same `.layout.rectilinear` submodules. I would instead expect to have the following in this file:

```python
from .connectivity.general import (
    create_directed_mesh_graph,
)
from .connectivity.triangular import (
    create_single_level_2d_mesh_graph,
    create_multirange_2d_mesh_graphs,
    create_single_level_2d_triangular_mesh_graph,
    create_flat_multiscale_from_triangular_coordinates,
)
from .layout.rectilinear import (
    create_multirange_2d_mesh_primitives,
    create_single_level_2d_mesh_primitive,
)
from .layout.triangular import (
    create_multirange_2d_triangular_mesh_primitives,
    create_single_level_2d_triangular_mesh_primitive,
)
```

and actually, I don't quite understand why we need the specialised functions for triangular meshes to construct digraphs (`create_single_level_2d_triangular_mesh_graph` and `create_flat_multiscale_from_triangular_coordinates`). So further adjustment to the above I would say the necessary functions and their modules should be:

```python
from .connectivity.general import (
    create_directed_mesh_graph,
)
from .connectivity.triangular import (
    create_single_level_2d_mesh_graph,  # works for rectilinear and triangular mesh layout
    create_multirange_2d_mesh_graphs,  # works for rectilinear and triangular mesh layout
)
from .layout.rectilinear import (
    create_multirange_2d_mesh_primitives,
    create_single_level_2d_mesh_primitive,
)
from .layout.triangular import (
    create_multirange_2d_triangular_mesh_primitives,
    create_single_level_2d_triangular_mesh_primitive,
)
```

what do you think?

src/weather_model_graphs/create/base.py
```diff
@@ -283,37 +309,103 @@ def create_all_graph_components(
...
+        ) or mesh_layout_kwargs.get("grid_spacing")
+        if mesh_node_spacing is None:
+            raise ValueError(
+                "mesh_layout='triangular' requires 'mesh_node_spacing' in "
+                "mesh_layout_kwargs (or 'mesh_node_distance' in "
+                "m2m_connectivity_kwargs for backward compatibility)."
+            )
+
+        if m2m_connectivity == "flat":
+            range_x, range_y = np.ptp(xy, axis=0)
```
leifdenby: these shouldn't be computed here, instead pass just `xy` to `create_single_level_2d_triangular_mesh_primitive` and compute nx_mesh etc inside

src/weather_model_graphs/create/base.py
```diff
@@ -283,37 +309,103 @@ def create_all_graph_components(
                     "max_num_refinement_levels"
                 ]
             G_mesh_coords = create_multirange_2d_mesh_primitives(**primitives_kwargs)
+
```
leifdenby: I think this can be combined with the `mesh_layout == 'rectilinear'` above to make it clearer. What about if you swap the `mesh_layout` and `m2m_connectivity` argument if-statement nesting and then for each `m2m_connectivity` pick the function to create the coordinate primitives based on the `mesh_layout` value?

src/weather_model_graphs/create/base.py
```diff
@@ -283,37 +309,103 @@ def create_all_graph_components(
...
-        graph_components["m2m"] = create_flat_multiscale_from_coordinates(
-            G_mesh_coords, **m2m_connectivity_kwargs
-        )
+        pattern = m2m_connectivity_kwargs.get("pattern")
+        if pattern is None:
+            pattern = "4-star" if mesh_layout == "triangular" else "8-star"
+        if mesh_layout == "triangular":
+            graph_components[
+                "m2m"
+            ] = create_flat_multiscale_from_triangular_coordinates(
```
leifdenby: as I wrote above I think we should try an generalise the `create_flat_multiscale_from_coordinates` function rather than writing a specalised one for `triangular` meshes

tests/test_mesh_layout.py
```diff
@@ -617,7 +617,7 @@ def test_unsupported_mesh_layout_raises(self):
             wmg.create.create_all_graph_components(
                 coords=xy,
                 m2m_connectivity="flat",
-                mesh_layout="triangular",
+                mesh_layout="hexagonal",
```
leifdenby: what does `hexagonal` refer to here?

leifdenby (changes requested):
Looking good :) I've made some suggestions and added some comments for discussion 🚀

docs/mesh_layout.ipynb
prajwal-tech07: Good catch - fixed, the intro now says mesh_layout is supported from v0.5.0.

src/weather_model_graphs/create/mesh/connectivity/triangular.py
prajwal-tech07: Agreed. For a triangular lattice the pattern argument is actually a no-op: the triangular primitive only ever contains the 6 lattice edges (all tagged adjacency_type="cardinal"), with no diagonal edges to include or exclude, so create_directed_mesh_graph returns the same 6-connectivity graph whether pattern is `4-star` or `8-star`. Rather than introduce a `6-star`, I think the cleanest fix is to drop the pattern argument from the triangular paths/docstrings entirely, since the connectivity is fully determined by the lattice. I'll fold this into the connectivity restructure below.

src/weather_model_graphs/create/mesh/__init__.py
prajwal-tech07: This matches how I think the split should work: 
`layout/` = coordinate primitives only (undirected nx.Graph), 
`connectivity/` = everything that turns primitives into a DiGraph. 

Concretely I'll:

- Remove `create_single_level_2d_triangular_mesh_graph` from `layout/triangular.py` - it's just primitive + `create_directed_mesh_graph`, so the generic path already covers it.
- Drop the layout → connectivity import so `layout/` no longer depends on `connectivity/`.
- Generalize `create_flat_multiscale_from_coordinates` to handle both layouts and delete `create_flat_multiscale_from_triangular_coordinates`.

One thing I'd like to confirm before touching the merged rectilinear path: the generic `create_flat_multiscale_from_coordinates` currently merges levels by grid-index arithmetic (it reshapes (i,j) node labels into a rectangle), which only works for rectilinear lattices. The triangular variant I wrote uses position-based KDTree matching, which is layout-agnostic. To generalize, I'd switch the rectilinear path to the same position-based matching. It should produce the same merged graph by construction (coarse nodes sit exactly on every Nth fine node), but it changes the merge mechanism on a path that's already merged and used by all archetypes - so I'll add a regression test asserting the existing rectilinear multiscale output is unchanged before/after. Does that approach sound good?

src/weather_model_graphs/create/base.py
prajwal-tech07: Yes Leif - I'll pass `xy` (and `mesh_node_spacing`) straight into `create_single_level_2d_triangular_mesh_primitive` and compute `nx/ny` inside it, so `base.py` doesn't need to know the lattice geometry.

src/weather_model_graphs/create/base.py
prajwal-tech07: Makes sense - I'll invert the nesting so `m2m_connectivity` is the outer branch and within each I pick the primitive-creation function from `mesh_layout`. That removes the duplicated `mesh_node_spacing` lookup/error-handling I currently have in both branches.

src/weather_model_graphs/create/base.py
prajwal-tech07: Yes Leif ,as covered in the `mesh/__init__.py` thread: I'll generalize the function and remove the triangular-specific one, pending your thumbs-up on the index→position-matching switch I described there.

tests/test_mesh_layout.py
prajwal-tech07: That's the negative test (`test_unsupported_mesh_layout_raises`) - `hexagonal` is just a stand-in for a layout we don't support yet, to confirm we raise `NotImplementedError`. I can rename it to something unambiguous like `nonexistent_layout` so it doesn't read like a real planned layout -want me to?

src/weather_model_graphs/create/mesh/layout/triangular.py
prajwal-tech07: Agreed, that import is the wrong direction. I'll move all DiGraph construction out of `layout/` so it only produces undirected coordinate primitives, and `connectivity/` owns everything that builds directed graphs. Full plan in the `mesh/__init__.py` thread below.

src/weather_model_graphs/create/mesh/layout/__init__.py
prajwal-tech07: Those were the two `networkx` helpers the layouts use under the hood - `networkx.grid_2d_graph` for the rectilinear layout and `networkx.triangular_lattice_graph` for the triangular one. You're right that naming the internal implementation functions in a user-facing module docstring is confusing/leaky, so I've reworded it to describe the layouts in plain terms (uniform rectangular grid vs regular equilateral triangular lattice) without referencing the networkx functions.

src/weather_model_graphs/create/mesh/__init__.py
leifdenby: hmm... is there a chance the KDTree matching based approach is slower? If so we could we maybe first see if the required attributes for the (i,j) node label are there and if not fall-back to the KDTree based approach?

tests/test_mesh_layout.py
leifdenby: ahh, yes maybe `nonexistang_layout` is better, just because one can construct hexagonals from triangles and so the "hexagonal" layout could be achieved by dropping every other edge in the triangle 6-star connection, so it is something we could have I think, kind of like a special case of triangular mesh, but it is not something we should add I think (it would also be ambiguous which edges to retain)

src/weather_model_graphs/create/mesh/__init__.py
prajwal-tech07: Good point - KDTree matching is `O(N log N)` per level vs the `O(N)` index arithmetic, so for large rectilinear meshes it'd be a bit slower (though small next to the rest of graph construction). I like your fallback: I'll keep `create_flat_multiscale_from_coordinates` on the fast index-based path when the primitives carry the integer `(i,j)` grid-index node labels (rectilinear), and fall back to position-based KDTree matching only when they don't (triangular). That also resolves the regression concern I raised - since rectilinear stays on the exact same code path, its output is unchanged by construction, and only triangular uses the KDTree branch. I'll key the decision off the node-label/attribute form, not the layout name, so it stays layout-agnostic.

tests/test_mesh_layout.py
prajwal-tech07: Sounds good - I'll rename the placeholder to `nonexistent_layout` in the unsupported-layout tests. And agreed: since "hexagonal" would just be a special case of the triangular mesh (dropping alternate edges from the 6-star, with ambiguity over which to keep), it makes sense not to add it as its own layout

src/weather_model_graphs/create/mesh/connectivity/triangular.py
prajwal-tech07: Done in f5b2d0e - dropped the `pattern` special-casing for triangular. Since the triangular primitive only has cardinal edges, `create_directed_mesh_graph` returns the same 6-connectivity regardless of pattern, so base.py now uses the default for both layout.

src/weather_model_graphs/create/mesh/layout/triangular.py
prajwal-tech07: Done in f5b2d0e - `layout/triangular.py` no longer imports from `connectivity/`, and the `create_single_level_2d_triangular_mesh_graph` DiGraph builder is removed (the generic create_directed_mesh_graph covers it). `layout/` now produces only undirected primitives.

src/weather_model_graphs/create/mesh/__init__.py
prajwal-tech07: Done in f5b2d0e. I went with your fallback idea: `create_flat_multiscale_from_coordinates` keeps the fast index-arithmetic merge for the rectilinear grid layout (so its output is unchanged by construction) and falls back to position-based KD-tree matching for layouts without a full grid (triangular). I detect the grid layout via the `"diagonal"` adjacency signature on the coordinate graph (pattern-independent) rather than the layout name, and the odd-`interlevel_refinement_factor` check now only guards the index path (the KD-tree path works for any factor). `create_flat_multiscale_from_triangular_coordinates` is removed, and I deleted the now-empty `connectivity/triangular.py`. One scope note: I left moving the rectilinear `create_single_level_2d_mesh_graph`/`create_multirange_2d_mesh_graphs` out of `layout/rectilinear.py` for a follow-up, since that touches the already-merged rectilinear path - happy to do it next if you'd like that to land here.

src/weather_model_graphs/create/base.py
prajwal-tech07: Done in f5b2d0e - `create_single_level_2d_triangular_mesh_primitive` now accepts `mesh_node_spacing` and computes `nx/ny` internally (mirroring the rectilinear primitive), so base.py just passes `xy` + spacing.

src/weather_model_graphs/create/base.py
prajwal-tech07: Done in f5b2d0e - inverted the nesting so `m2m_connectivity` is the outer branch (single- vs multi-level) and the primitive function is selected by `mesh_layout` inside. The duplicated `mesh_node_spacing` lookup/validation is now done once.

src/weather_model_graphs/create/base.py
prajwal-tech07: Done in f5b2d0e - covered in the `mesh/__init__.py` thread: the function is generalized (index path for rectilinear, KD-tree fallback otherwise) and the triangular-specific one is removed.

tests/test_mesh_layout.py
prajwal-tech07: Done in f5b2d0e - renamed the placeholder to `nonexistent_layout` in the unsupported-layout tests.

src/weather_model_graphs/create/base.py
```diff
@@ -28,10 +29,14 @@
     create_flat_singlescale_from_coordinates,
 )
 from .mesh.connectivity.hierarchical import create_hierarchical_from_coordinates
-from .mesh.coords import (
+from .mesh.layout.rectilinear import (
     create_multirange_2d_mesh_primitives,
     create_single_level_2d_mesh_primitive,
 )
+from .mesh.layout.triangular import (
+    create_multirange_2d_triangular_mesh_primitives,
```
leifdenby: we don't have to do this, but what do you think about changing the names of the functions here to avoid the `_triangular` part and just rename during import, like so:

```suggestion
    create_multirange_2d_mesh_primitives as create_multirange_2d_triangular_mesh_primitives,
```

we could do the same with the rectilinear mesh primitives the import above too. Might make it easier to read the code later (since in this module the rectilinear functions don't have `rectilinear` in their name, but maybe we should do a `import x as y` here too)

src/weather_model_graphs/create/base.py
```diff
@@ -255,64 +264,101 @@ def create_all_graph_components(
...
-            if "refinement_factor" in mesh_layout_kwargs:
-                primitives_kwargs["interlevel_refinement_factor"] = mesh_layout_kwargs[
-                    "refinement_factor"
-                ]
-            if "max_num_refinement_levels" in mesh_layout_kwargs:
-                primitives_kwargs["max_num_levels"] = mesh_layout_kwargs[
-                    "max_num_refinement_levels"
-                ]
-            G_mesh_coords = create_multirange_2d_mesh_primitives(**primitives_kwargs)
+        else:  # triangular
```
leifdenby: i think we should have `elif == 'triangular'` here instead, and then raise a not implemented exception on any other value

src/weather_model_graphs/create/base.py
```diff
@@ -255,64 +264,101 @@ def create_all_graph_components(
...
+            primitives_kwargs["interlevel_refinement_factor"] = mesh_layout_kwargs[
+                "refinement_factor"
+            ]
+        if "max_num_refinement_levels" in mesh_layout_kwargs:
+            primitives_kwargs["max_num_levels"] = mesh_layout_kwargs[
+                "max_num_refinement_levels"
+            ]
+        if mesh_layout == "rectilinear":
+            G_mesh_coords = create_multirange_2d_mesh_primitives(**primitives_kwargs)
+        else:  # triangular
```
leifdenby: should also use `elif`

tests/test_triangular_mesh.py
```diff
@@ -0,0 +1,995 @@
...
+Tests verify:
+1. Primitive creation (node count, positions, adjacency_type, type attrs)
+2. Single-level directed graph (bidirectional edges, len/vdiff attrs, 6-connectivity)
+3. Multirange primitive creation
+4. Flat multiscale mesh graph (position-based merging via two-step API)
+5. Hierarchical mesh graph (triangular primitives + generic hierarchical connectivity)
+6. Integration through create_all_graph_components for all m2m_connectivity modes
+7. Edge cases (zero nodes, single-level hierarchical)
+8. Numerical correctness (len symmetry, vdiff reciprocity)
+9. Pattern equivalence (4-star == 8-star for triangular)
```
leifdenby: does this makes sense? I thought there was only 6-star for triangular mesh coordinate layouts?

tests/test_triangular_mesh.py
```diff
@@ -0,0 +1,995 @@
...
+Tests for mesh_layout="triangular" support (Issue #80).
+
+Tests verify:
+1. Primitive creation (node count, positions, adjacency_type, type attrs)
+2. Single-level directed graph (bidirectional edges, len/vdiff attrs, 6-connectivity)
+3. Multirange primitive creation
+4. Flat multiscale mesh graph (position-based merging via two-step API)
+5. Hierarchical mesh graph (triangular primitives + generic hierarchical connectivity)
+6. Integration through create_all_graph_components for all m2m_connectivity modes
+7. Edge cases (zero nodes, single-level hierarchical)
```
leifdenby, line 11: why would there be zero nodes?

tests/test_triangular_mesh.py
```diff
@@ -0,0 +1,995 @@
...
+    create_hierarchical_from_coordinates,
+)
+from weather_model_graphs.create.mesh.layout.triangular import (
+    create_multirange_2d_triangular_mesh_primitives,
+    create_single_level_2d_triangular_mesh_primitive,
+)
+
+
+# Test helpers exercising the equivalent generic two-step API after the
+# triangular-specific convenience functions were removed (PR #92 review).
```
leifdenby: I don't think we need this, do we?

tests/test_triangular_mesh.py
```diff
@@ -0,0 +1,995 @@
...
+def create_single_level_2d_triangular_mesh_graph(xy, nx, ny):
+    """Directed triangular mesh graph via primitive + generic connectivity."""
+    return create_directed_mesh_graph(
+        create_single_level_2d_triangular_mesh_primitive(xy, nx, ny)
+    )
+
+
+def create_flat_multiscale_from_triangular_coordinates(G_coords_list, **kwargs):
```
leifdenby: why do we need this?

tests/test_triangular_mesh.py
```diff
@@ -0,0 +1,995 @@
...
+        G_coords = create_single_level_2d_triangular_mesh_primitive(
+            xy_small, nx=5, ny=5
+        )
+        G = create_directed_mesh_graph(G_coords, pattern="4-star")
+        for u, v in G.edges():
+            if G.has_edge(v, u):
+                np.testing.assert_allclose(G[u][v]["len"], G[v][u]["len"], atol=1e-10)
+
+    def test_vdiff_reciprocity(self, xy_small):
+        """vdiff(uΓåÆv) should equal -vdiff(vΓåÆu)."""
```
leifdenby: something funny with the utf characters here I htink :)

leifdenby (changes requested):
Looks great! Only a few small suggestions from me :)

src/weather_model_graphs/create/base.py
prajwal-tech07: Done - switched to `elif mesh_layout == "triangular"` and raise `NotImplementedError` for any other value.

src/weather_model_graphs/create/base.py
prajwal-tech07: Done - same `elif` + `NotImplementedError` here.

tests/test_triangular_mesh.py
prajwal-tech07: The `pattern` (4-star/8-star) selects edges by `adjacency_type`: 4-star keeps cardinal, 8-star keeps cardinal + diagonal. The triangular lattice only has cardinal edges, so both patterns give the same graph - that's what this test checks. The "6" you're thinking of is the node degree (6 neighbours per interior node), which is separate from the pattern name. I've clarified the docstring to spell this out.

tests/test_triangular_mesh.py
prajwal-tech07: It just guards the degenerate input - if someone passes `nx=0` or `ny=0` we raise a clear error instead of silently returning an empty mesh. Happy to drop it if you'd rather not test that path.

tests/test_triangular_mesh.py
prajwal-tech07: Removed - inlined the helpers so the tests call the generic two-step API (`create_directed_mesh_graph(create_single_level_2d_..._primitive(...))`) directly.

tests/test_triangular_mesh.py
prajwal-tech07: Removed - it was a pure pass-through, so I replaced the call sites with `create_flat_multiscale_from_coordinates` directly.

tests/test_triangular_mesh.py
prajwal-tech07: Fixed - the arrows had been mangled by an encoding issue; replaced them with ASCII `->`.

src/weather_model_graphs/create/base.py
prajwal-tech07: Done - renamed the triangular primitives to drop `_triangular` and aliased them on import. Did the same for the rectilinear ones too, so both layouts are symmetric.

prajwal-tech07:
Thanks for the review! Addressed all the suggestions in fc01d66: renamed the layout primitives (+ alias on import), switched both dispatch branches to `elif` + `NotImplementedError`, inlined/removed the two test helpers, clarified the 4-star/8-star docstring, and fixed the mangled UTF arrows. Left one question open on the zero-nodes test above.Full suite passes locally.

tests/test_triangular_mesh.py
leifdenby: ok, but I don't understand why `pattern=='4-star'` or `pattern=='8-star'` even makes sense as an argument :) Can you point me to where that is?

tests/test_triangular_mesh.py
leifdenby: ok that is fine :)

tests/test_triangular_mesh.py
prajwal-tech07: Yeah, I got a bit confused about this earlier but it's clear now. You're right. The pattern is handled in `create_directed_mesh_graph` - `src/weather_model_graphs/create/mesh/connectivity/general.py`, around lines 41–54: `4-star` keeps only cardinal edges, `8-star` keeps `cardinal + diagonal`. That's really a rectilinear distinction. The triangular primitive tags every edge as `cardinal`, so the pattern doesn't change anything for triangular - `4-star` and `8-star` come out identical. So that test wasn't really checking anything meaningful. I've removed it and pushed.

prajwal-tech07:
Pushed 09e1a50, which adds the `MESH_LAYOUT_OPTIONS` constant we discussed in mllam/neural-lam#661.

**What it does**
- Defines `MESH_LAYOUT_OPTIONS` in `create/mesh/layout/__init__.py`, next to the layout modules themselves, so the package that owns the layouts owns the list
- `create_all_graph_components` now validates against it, replacing the three hardcoded `("rectilinear", "triangular")` tuples
- Re-exported so it resolves as `wmg.create.MESH_LAYOUT_OPTIONS`, as you suggested

**Why in this PR rather than a separate one** - the `layout/` package only exists here, so a standalone PR against `main` would have to either recreate the package or ship a constant listing only `rectilinear`, which this PR would then have to move and extend anyway. It also felt like the natural home: one layout doesn't need a registry, two do.

**Notes**
- The rejection message is now generated from the constant but renders identically to before, so the existing tests are untouched
- New `TestMeshLayoutOptions` parametrises over the constant × all three `m2m_connectivity` modes. That's the drift guard: adding a name to `MESH_LAYOUT_OPTIONS` without wiring it into the dispatch now fails loudly instead of advertising something unusable
- I used a tuple rather than a list, so consumers can't mutate a public constant. Happy to switch to a list if you'd prefer.

Full suite green (306 passed, up from 298 - the 8 new tests), pre-commit clean.

This unblocks the `--mesh_layout` CLI work in mllam/neural-lam#661, and means #91 and #76 can each add their layout name in one place.

prajwal-tech07:
Pushed a29fa97, a bugfix I hit while adding the tests above.

`create_multirange_2d_mesh_primitives` took `max_num_levels` as a **required positional argument**, but `create_all_graph_components` only forwards it when `max_num_refinement_levels` is present in `mesh_layout_kwargs`. So this raised a `TypeError`, despite the argument being documented as optional:

```python
wmg.create.create_all_graph_components(
    coords=xy,
    m2m_connectivity="flat_multiscale",      # or "hierarchical"
    mesh_layout="rectilinear",
    mesh_layout_kwargs=dict(mesh_node_spacing=3),   # no max_num_refinement_levels
    g2m_connectivity="nearest_neighbour",
    m2g_connectivity="nearest_neighbour",
)
# TypeError: create_multirange_2d_mesh_primitives() missing 1 required
# positional argument: 'max_num_levels'
```

**This affects released v0.4.0 on `main` too**, not just this branch - I verified it against a clean checkout of `main`. It hasn't been hit because all three archetypes always pass the argument, so only direct `create_all_graph_components` calls are affected.

I've fixed it here rather than in a separate PR because this branch introduces a second copy of the same signature in the triangular layout, so it seemed better not to propagate it. The function body already treats a falsy value as "no limit", so only the signature needed changing: `xy` moves to the front and the rest become keyword-only, matching `create_single_level_2d_mesh_primitive` in the same modules. Every call site already used keywords, and a stray positional call now fails loudly instead of silently binding to the wrong parameter. Happy to split it out if you'd rather keep this PR to the layout work.

---

**A separate question about the triangular layout that I'd like your view on.**

While checking how the layouts behave, I found that the lattice this PR builds isn't equilateral. `layout/triangular.py` computes `scale_x` and `scale_y` independently to fit the domain, which stretches it. On a 100×100 grid with `mesh_node_spacing=5.0` I measure two distinct m2m edge lengths, **6.2272 and 9.0** - an equilateral lattice should have exactly one.

To be fair to the existing design, rectilinear does the same thing: on a 200×60 domain I get `dx=5.1026, dy=5.3636`. So "stretch to fit the domain" is the established convention from #81 and this PR is following it. But it does mean the docstring's "equilateral-triangle lattice" and "more isotropic message passing" aren't accurate as written - and triangular is stretched even on a *square* domain, because the raw lattice aspect never matches the domain aspect.

So: should `mesh_node_spacing` mean **"stretch the lattice to fit the domain"** (current behaviour, consistent with rectilinear) or **"the actual distance between mesh nodes"** (lattice sized to cover the domain instead)?

If I fix the scale to the requested spacing and size the lattice to cover, I get a single edge length of exactly 5.0.

I'd lean towards the second, since it makes the isotropy claim true and makes the `g2m` radius predictable per layout (see my comment on mllam/neural-lam#661). But it changes what `mesh_node_spacing` means, and #76 will hit the identical question for icosahedral, so it's probably worth deciding once across all layouts. Happy to go either way - I'll update the docstring to match whichever we pick.

leifdenby:
> To be fair to the existing design, rectilinear does the same thing: on a 200×60 domain I get `dx=5.1026, dy=5.3636`. So "stretch to fit the domain" is the established convention from #81 and this PR is following it. But it does mean the docstring's "equilateral-triangle lattice" and "more isotropic message passing" aren't accurate as written - and triangular is stretched even on a _square_ domain, because the raw lattice aspect never matches the domain aspect.

Oh, I hadn't realised this! Does that mean we changed the default behaviour here with #81? I would expect for the rectilinear layout that the mesh spacing is indeed equal, and the same for the triangular mesh we are introducing here.

prajwal-tech07:
No, this predates #81. The same scaling was already in mesh.py before it, #81 just moved the code into the layout module and renamed the argument.

Rectilinear looks fine to me. The only thing that shifts the spacing there is the node count getting rounded down, so if the domain divides evenly you get exactly what you asked for, and if it doesn't it's off by less than one node.

Triangular is where it's actually broken. The lattice gets stretched by different amounts in x and y to fit the domain, so the triangles come out skewed no matter what domain you give it. That's why the equilateral and isotropy claims in the
docstring don't hold.

I've changed it to scale both directions by the same factor, so the triangles stay equilateral and mesh_node_spacing is the real distance between nodes now. Side effect is that the lattice doesn't line up with the domain edges anymore, so I sized it to cover the domain, which means the outermost nodes can sit slightly outside it.

While doing that I ran into something else. Because each level was being stretched on its own, coarse level nodes never landed on fine level nodes, and multiscale merges levels by matching positions. So flat_multiscale on a triangular mesh was
giving one disconnected component per level instead of an actual multiscale graph. Anchoring all the levels to the same origin fixes that too. I added a connectivity test, the existing ones only checked node counts which is why it slipped through.

leifdenby:
Based on looking through the examples in the `mesh_layout.ipynb` notebook together just now there does seem to be an issue with the edge lengths for the triangular mesh. The `networkx` function used to generate the coordinates claims that it should be a equilateral triangular layout (of edge length 1), but something is going wrong.

billede

docs/mesh_layout.ipynb
leifdenby, line 1: Make sure to plot all graphs with `ax.set_aspect(1)` :)

leifdenby: I've committed my notebook that we worked on during out chat here in case that is useful to you: https://github.com/leifdenby/weather-model-graphs/tree/review/feat/triangular-mesh-layout

prajwal-tech07: Done in 577e3c8, three of the plots were missing it. Also added a print of the min/max edge length under the triangular example so it stays visible in the docs.

prajwal-tech07:
Yeah, what we were looking at in the meeting was actually a bit wrong , your branch is based on the June notebook commit, so it was still running the old scaling. I'd pushed the fix for this in `2bc8d58` but hadn't mentioned it, sorry.

The old code stretched x and y separately, which is why the edges split into two groups. Running your notebook against that old code I get 3.031 and 3.429, which matches the colours in your screenshot. On the current branch the spread is gone:

    np.ptp(edge_lengths)   ->   3.1e-15

One heads up if you rerun the cell: `np.unique` will still show a few entries that all print as `3.`. That's just the last couple of bits from the sqrt in the diagonal length (2.9999999999999987, 3.0, 3.0000000000000018), so `np.ptp` or rounding is the clearer check. Before the fix that same spread was 0.397, about 13% of the requested spacing.

Thanks for pushing your notebook, made this easy to check against.

PR #111, opened 2026-03-23
Improve specificity of decode_mask test by validating m2g edges

bhawna346:
This PR improves the specificity of the decode_mask test by focusing on the relevant decoding (m2g) edges instead of checking the total number of edges.

Previously, the test verified overall edge count changes, which could lead to vague or misleading validation. This update ensures that only the m2g edges are evaluated, making the test more aligned with the intended behavior and more robust to future changes.

Addresses #103


zweihuehner:
Duplicate of #104

PR #117, opened 2026-03-25
[Feat] runtime scaling and flamegraph benchmark CLIs

leifdenby:
## Describe your changes

It is clear that the current implementation of `weather-model-graphs` can lead to quite long runtime with the number of grid points in dataset that people have been working (for example @observingClouds and @joeloskarsson have seen this). 

To try and quantify the scaling and to try an identify where improvements could be made this PR adds two CLI scripts in `tests/benchmarks/`:

1. A runtime-scaling script (`uv run python -m tests.benchmarks.graph_creation_scaling`) that executes graph creation over progressively larger input grid sizes. It generates a plot of execution time vs. number of input nodes alongside a linear $O(N)$ reference line, allowing us to easily visualize how the algorithm's performance degrades as coordinate sizes grow.
2. A hierarchical profiling script (`uv run python -m tests.benchmarks.graph_creation_flamegraph`) powered by `pyinstrument`. It generates and automatically serves an interactive HTML call-stack flamegraph, making it simple to pinpoint exactly which internal functions (e.g., `scipy` KDTree queries, or `networkx` node iteration) are responsible for performance bottlenecks.

I thought these CLI tools could provide a baseline for future optimisation work - we could even eventually add CI tests checking that the execution time doesn't drastically blow up as we work on the codebase.

- `pyinstrument` has been added to the `dev` dependencies in `pyproject.toml` to support the interactive flamegraphs.

### Runtime scaling benchmark

| `flat` | `flat_multiscale` | `hierarchical` |
| :---: | :---: | :---: |
| scaling_keisler | scaling_graphcast | scaling_oskarsson_hierarchical |

The good news is that **runtime scales roughly linearly with number of input coord positions**, of course this will vary with coordinate density, layout etc, this is for regular rectilinearly laid out input grid nodes. 

The bad news is that for order $10^5$ nodes the runtime (at least on my laptop) is order `~6s`, about $10^6$ we are looking at runtime of a minute, but another order of magnitude $10^7$ would be an hour.

But, I think there are some clear optimisations we can do, which brings me to the flamegraphs...

### Flamegraph profiling script

I'm attaching here two screenshots taken from the browser window that opens when you run the flamegraph benchmark script. This is the [pyinstrument](https://github.com/joerick/pyinstrument) interface:

| Call Stack | Timeline |
| :---: | :---: |
| Screenshot 2026-03-25 at 16 40 00 | Screenshot 2026-03-25 at 16 40 06 |

I've included [the `.html`-profile result page too](https://github.com/user-attachments/files/26248048/pyinstrument_profile.html)) in case someone wants to look further into this example.

It looks to me (as I expected) that there a repeated calls inside `connect_nodes_across_graphs` which could be vectorized quite easily where we would query many points at once and create all the edges with one call on the `networkx.DiGraph` object being worked on. However, the composing of subgraphs together and relabelling them at the end appears to be quite costly too, the former might require a different datastructure that `nx.DiGraphs` which might require a rethink, for the latter maybe we can do something smarter (it shouldn't be that costly I don't think...)

It is not my aim to start a long discussion here about optimisation that could be done. I suggest we open a separate issue for that. Instead, I would simply like to add these tools so that we can establish a way to investigate the issue and plan a course of action.

## Issue Link

This PR supplements the CPU vs GPU based benchmark that https://github.com/mllam/weather-model-graphs/pull/62 is introducing.

## Type of change

- [ ] I have performed a self-review of my code
- [ ] For any new/modified functions/classes I have added docstrings that clearly describe its purpose, expected inputs and returned values
- [ ] I have placed in-line comments to clarify the intent of any hard-to-understand passages of my code
- [ ] I have given the PR a name that clearly describes the change, written in imperative form ([context](https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form)).

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


leifdenby:
Since this is completely orthogonal to the existing functionality (i.e. doesn't break anything) and already being built on by others #137, #138, #140, and #142 I am going to move this forward to the next release and merge it in. We briefly discussed this in a previous dev meeting and I just hadn't had time to do anything about it yet

prajwal-tech07:
Reviewed this and ran both scripts locally against current main (merged it in first since the branch is from before #81). Really useful addition - having a baseline for the scaling/profiling work is great, and the code is clean and well-documented. A few notes, all non-blocking:

One thing worth handling - hierarchical at small grid sizes. Running the scaling script with a small `--min-N` and `--archetype oskarsson_hierarchical` raises `ValueError: At least two mesh levels are required for hierarchical mesh graph` (it needs enough grid points to form ≥2 levels). The default `--min-N 50` works fine, but a quick small-N run crashes with an error that isn't obviously about grid size. Might be worth catching it and skipping/warning with a benchmark-friendly message, or noting the minimum grid size for hierarchical in the README.

Minor suggestions:

- Both scripts use `time.time()` for timing - `time.perf_counter()` would be slightly better for benchmarks (monotonic, higher resolution, unaffected by system clock changes).
- The scaling script times a single run per N. A couple of repeats taking the min/median would give less noisy points (and a steadier O(N) reference, since it's anchored to the first point). Probably future-work given this is meant as a baseline.

**What I confirmed works:** keisler and graphcast scaling plots both generate correctly, and the flamegraph script works in both `--save-flamegraph` and `--console` modes (the HTML output is valid). Nice that the server binds to `127.0.0.1` on a free port and cleans up properly.

Overall looks good to me - solid, low-risk, and genuinely helpful.

PR #123, opened 2026-03-28
to_neural_lam() to wmg.save for neural-lam tensor-on-disk format

prajwal-tech07:
## Describe your changes

Add a `to_neural_lam()` function to `wmg.save` that serializes graph components (from `wmg.create.archetype.*(..., return_components=True)`) to the tensor-on-disk `.pt` format expected by `neural_lam.utils.load_graph()`.

This is the first step in bridging `weather-model-graphs` and `neural-lam` — putting the serialization logic in wmg so that neural-lam's CLI can remain a thin wrapper.

**What it does:**
- Accepts the `{"g2m", "m2m", "m2g"}` dict returned by any archetype with `return_components=True`
- Writes 7 core `.pt` files (g2m/m2g as single tensors, m2m/mesh_features as lists)
- Writes 4 additional files for hierarchical graphs (mesh_up/mesh_down edge_index and features)
- Edge features are written **raw** (unnormalized) — neural-lam normalizes at load time
- Mesh node positions are pre-normalized by `max(abs(pos))`, matching the existing neural-lam convention

**Supports all three archetypes:**
- Keisler (flat single-scale)
- GraphCast (flat multiscale)
- Oskarsson (hierarchical)

Also adds two internal helpers (`_graph_to_edge_tensors`, `_graph_to_node_features`) that cleanly strip heterogeneous node/edge attributes before PyG conversion.

**Tests:** 22 new tests across 4 test classes covering all archetypes, shape/dtype validation, normalization checks, and edge cases.

No new dependencies — uses the existing `[pytorch]` optional extras (torch, torch-geometric).

## Issue Link

Refs mllam/neural-lam#384

## Type of change

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested

## Checklist for assignee

- [ ] the tests pass


prajwal-tech07:
Fixed the pre-commit linting failure in a1ec5b5:


- Applied black formatting to `save.py` and `test_save.py`
- Removed unused `import numpy as np` from `test_save.py` (flake8 F401)

All pre-commit hooks and tests (24/24) pass locally now. The CI should be green on the next run.

prajwal-tech07:
Reverted the strict `torch-geometric==2.3.1`) pin back to the flexible >=2.5.3 requirement (matching upstream main) in ef36af0. This way we don't restrict users to a specific version, consistent with how `neural-lam` handles version requirements.

All pre-commit hooks and tests (24/24) pass.

prajwal-tech07:
Renamed to_neural_lam → `to_torch_tensors_on_disk` as suggested — this better describes what the function actually does (save torch tensors to disk) without namespacing it to neural-lam specifically. All test references and class names updated accordingly. All 24 tests pass.

The corresponding neural-lam PR (mllam/neural-lam#596) has been updated to use the new function name as well.

prajwal-tech07:
Hi @leifdenby! I've finished the implementation for this PR. In addition to `to_torch_tensors_on_disk()`, I've also added a `validate_neural_lam_graph_dir()` function that verifies a saved graph directory matches the neural-lam tensor format spec (correct dtypes, shapes, normalized mesh positions, non-negative indices, hierarchical file presence). Ready for review!

src/weather_model_graphs/save.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+    features : torch.Tensor
+        Shape (num_edges, num_feature_cols), dtype float32. With default
+        features this is (num_edges, 3) for [len, vdiff_x, vdiff_y].
+    """
+    if not HAS_PYG:
+        raise RuntimeError(
+            "install weather-model-graphs[pytorch] to enable writing to torch files"
+        )
+
+    if edge_features is None:
```
leifdenby: rather than having `None` be the default argument and then replacing that value here we should expose the default values in the function arguments

src/weather_model_graphs/save.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+        edge_features = ["len", "vdiff"]
+
+    # Deterministic edge ordering: sort by (sender index, receiver index)
```
leifdenby: why do we need to do this? Add comment to explain

src/weather_model_graphs/save/neural_lam/torch_tensors.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+    if edge_features is None:
+        edge_features = ["len", "vdiff"]
+
+    # Deterministic edge ordering: sort by (sender index, receiver index)
+    edges = sorted(
+        graph.edges(data=True),
+        key=lambda e: (sender_map[e[0]], receiver_map[e[1]]),
+    )
+
+    if len(edges) == 0:
```
leifdenby, line 91: under what situations might this arise? As in, why do we need to be able to handle this case? I can't quite see why we'd want any nodes to be completely un-connected which I think having no edges would imply

src/weather_model_graphs/save/neural_lam/torch_tensors.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+            [sender_map[u] for u, _, _ in edges],
+            [receiver_map[v] for _, v, _ in edges],
+        ],
+        dtype=torch.int64,
+    )
+
+    rows = []
+    for _, _, data in edges:
+        vals = [
+            np.atleast_1d(np.asarray(data[f], dtype=np.float64)) for f in edge_features
```
leifdenby, line 112: why is `np.atleast_1d` needed here?

src/weather_model_graphs/save.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
     logger.info(f"Saved node features {node_features} to {fp_node_features}.")
 
 
+def _graph_to_edge_tensors(graph, sender_map, receiver_map, edge_features=None):
```
leifdenby: missing type annotations for arguments and return

src/weather_model_graphs/save.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+    torch.Tensor
+        Shape (len(labels), num_feature_cols), dtype float32. With default
+        features this is (num_nodes, 2) for [pos_x, pos_y].
+    """
+    if not HAS_PYG:
+        raise RuntimeError(
+            "install weather-model-graphs[pytorch] to enable writing to torch files"
+        )
+
+    if node_features is None:
```
leifdenby: as above, defaults should be exposed in function arguments, not replacing `None` placeholder

src/weather_model_graphs/save.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+    for n in labels:
+        data = graph.nodes[n]
+        vals = [
+            np.atleast_1d(np.asarray(data[f], dtype=np.float64)) for f in node_features
+        ]
+        rows.append(np.concatenate(vals))
+    return torch.tensor(np.stack(rows), dtype=torch.float32)
+
+
+GRAPH_STORAGE_SPEC_VERSION = "0.1.0"
```
leifdenby: put at top of file, constants should be written part way down a file

src/weather_model_graphs/save.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+):
+    """
+    Save graph components to the neural-lam tensor-on-disk format.
+
+    Takes graph components as returned by
+    ``wmg.create.archetype.*(..., return_components=True)`` and writes
+    ``.pt`` files conforming to the neural-lam graph storage specification
+    (version 0.1.0), as expected by ``neural_lam.utils.load_graph()``.
+
+    Both edge features and mesh node features (positions) are written
```
leifdenby: Why do we need this detail here?

src/weather_model_graphs/save/neural_lam/torch_tensors.py
```diff
@@ -158,6 +159,400 @@ def _concat_pyg_features(
...
+GRAPH_STORAGE_SPEC_VERSION = "0.1.0"
+
+
+def to_torch_tensors_on_disk(
+    graph_components: Dict[str, networkx.DiGraph],
+    output_directory: str,
+    hierarchical: bool = False,
+):
+    """
+    Save graph components to the neural-lam tensor-on-disk format.
```
leifdenby, line 281: I think we need some more details here about what the steps within this function actually are. Can you make a numbered list here in the docstring to explain the steps of the function?

First we need to explain what "graph components" are, and then explain the steps of the code below

leifdenby (changes requested):
With the graph spec merged in `neural-lam` (https://github.com/mllam/neural-lam/pull/323) we can proceed on getting this finished of and merged in :)

I've left some comments. Can you address these and I will do a second review? Thanks!

src/weather_model_graphs/save.py
prajwal-tech07: Added a comment in 23fb305. Sorting makes the on-disk edge order deterministic and independent of networkx's iteration order, so the same graph always serialises to the same `edge_index`/feature ordering - otherwise you get equivalent-but-different orderings between runs, which breaks reproducibility and makes diffing graphs harder.

src/weather_model_graphs/save/neural_lam/torch_tensors.py
prajwal-tech07: This guards a subgraph with no edges - e.g. an inter-level up/down direction with no connections for a given level pair - since `np.stack` raises on an empty list, so I return correctly-shaped empty tensors instead. Added a comment in 23fb305. In the standard archetypes every adjacent level does have up/down edges, so it's mostly defensive - happy to drop it if you'd prefer.

src/weather_model_graphs/save/neural_lam/torch_tensors.py
prajwal-tech07: Added a comment in 23fb305 - `"len"` is a scalar attribute while `"vdiff"` is a 2-/3-vector, and `np.concatenate` needs 1-D inputs, so `atleast_1d` promotes the scalar to a 1-element array so they concatenate into a single row.

src/weather_model_graphs/save.py
prajwal-tech07: Done in 23fb305, same as the edge case - exposed via `DEFAULT_NODE_FEATURES = ("pos",)` in the signature instead of the `None` placeholder.

src/weather_model_graphs/save.py
prajwal-tech07: Moved `GRAPH_STORAGE_SPEC_VERSION` (and the new default-feature constants) to the top of the file in 23fb305.

src/weather_model_graphs/save.py
prajwal-tech07: Trimmed it in 23fb305 - folded the raw/unnormalized note into the numbered steps rather than leading the docstring with it, so it sits next to the step that writes the features.

src/weather_model_graphs/save/neural_lam/torch_tensors.py
prajwal-tech07: Rewrote the docstring in 23fb305: it now starts by explaining what a graph component is (g2m/m2m/m2g and the node attributes they carry), then gives a numbered list of the steps the function performs.

src/weather_model_graphs/save.py
prajwal-tech07: Added type annotations for the arguments and return in 23fb305 — also did the same on `_node_features_from_labels` for consistency.

src/weather_model_graphs/save.py
prajwal-tech07: Done in 23fb305 the default is now a module-level `DEFAULT_EDGE_FEATURES = ("len", "vdiff")` used directly in the signature, instead of the `None` placeholder (a tuple so it's safe as an argument default).

src/weather_model_graphs/save/neural_lam/torch_tensors.py
leifdenby: we should raise an exception here instead, no edge sets should be empty

leifdenby:
This is looking great!

Things to check:
- don't use default arguments of value `None` where that value is overwritten locally in the function
- ensure all functions have type annotations on arguments and return values
- ensure all functions have numpy style docstrings

src/weather_model_graphs/save/neural_lam/torch_tensors.py
```diff
@@ -158,6 +170,435 @@ def _concat_pyg_features(
     logger.info(f"Saved node features {node_features} to {fp_node_features}.")
 
 
+def _graph_to_edge_tensors(
```
leifdenby, line 21: judging from the number of if-statements in this function I feel like we should probably implement some tests. I just had Claude have a look and it suggests these tests below, what do you think?

- `HAS_PYG == False`
  - Raises `RuntimeError` before doing anything else.

- Graph with zero edges (**I think we should raise an exception here instead**)
  - Returns `edge_index` shape `(2, 0)` and `features` shape `(0, 3)`.
  - Verify this works for any `sender_map` / `receiver_map` when no edges exist.

- Graph with edges present
  - Returns edges sorted by `(sender_map[u], receiver_map[v])`, not NetworkX iteration order.
  - Confirm deterministic ordering across differently ordered input graphs.

- Missing node in `sender_map` or `receiver_map`
  - Should raise `KeyError` during sorting or tensor construction.
  - Test both missing sender-side and receiver-side labels.

- Missing edge attribute in `edge_features`
  - Should raise `KeyError` when indexing `data[f]`.

- Mixed scalar/vector edge attributes
  - Default case: scalar `len` plus vector `vdiff` should concatenate into 3 columns.
  - Test with 2D and 3D `vdiff` if supported by the input graphs.

- Edge attribute shape mismatch across edges
  - One edge has `vdiff` length 2, another length 3, or one attribute is scalar when others are vectors.
  - Should fail in `np.concatenate` or `np.stack`.

- Empty `edge_features` tuple/list
  - `rows` becomes arrays of length 0, so output shape may be `(num_edges, 0)`.
  - Worth testing because it’s a corner case not guarded by an `if`.

- Non-numeric edge attribute values
  - `np.asarray(..., dtype=np.float64)` should fail if an attribute is a string/object.

- Edge data stored as numpy arrays / lists / Python scalars
  - Verify all are normalized correctly by `np.atleast_1d`.

- Single-edge graph
  - Confirms the non-empty path works with minimal input.

leifdenby (commented):
I feel like `weather_model_graphs.save` is getting quite large. Should we maybe move it into a submodule called `weather_model_graphs.save.neural_lam.torch_tensors`? I think we should move `wmg.save.to_pyg()` to `weather_model_graphs.save.neural_lam.deprecated` and add a warning when people call this function and suggest they call the new `to_torch_tensors_on_disk()` and make it clear this function will not be maintained going forward

The functionality that is general between the two output methods could go in `weather_model_graphs.save.base`.

What do you think?

src/weather_model_graphs/save/neural_lam/torch_tensors.py
leifdenby: Also the default arguments that define which named attributes to output should use those global variables we have in https://github.com/prajwal-tech07/weather-model-graphs/blob/issue-384/to-neural-lam/src/weather_model_graphs/save.py#L32 I think. What do you think?

prajwal-tech07: Thanks Leif, addressed everything from this round, pushed in 6de5776 (+ db925d2 merging in latest main, 51b3e71 small test fix after that merge).

**On the restructure question**: yes, agree, done. save is now a package laid out as you suggested:

- `save/base.py` - the shared bits: `HAS_PYG` detection, the `DEFAULT_EDGE_FEATURES` / `DEFAULT_NODE_FEATURES` constants, and `to_pickle`
- `save/neural_lam/torch_tensors.py` - `to_torch_tensors_on_disk` and its helpers
- `save/neural_lam/deprecated.py` - `to_pyg`, which now emits a `DeprecationWarning` telling people to use `to_torch_tensors_on_disk` and is documented as no longer maintained

`save/__init__.py` re-exports all of them, so existing `wmg.save.to_pyg(...)` / `wmg.save.to_torch_tensors_on_disk(...)` call sites keep working unchanged.


**Empty edge set:** now raises `ValueError` instead of returning empty tensors. I also checked empirically that none of the keisler/graphcast/hierarchical archetypes ever produce an empty edge set (including the inter-level up/down subgraphs), so the raise doesn't affect any normal path.

**Tests for `_graph_to_edge_tensors`:** added, covering the list: the no-pytorch RuntimeError guard, empty edge set raising, shapes/dtypes, deterministic ordering across differently-ordered input graphs, missing node in sender/receiver map (KeyError), missing edge attribute (KeyError), inconsistent vdiff lengths and non-numeric attributes (both ValueError), and the single-edge case. The only one I left out is the "empty edge_features tuple" case, since with the defaults now exposed in the signature that's not a reachable path in normal use. Also added a test that to_pyg emits the deprecation warning.

**Defaults from the globals:** done. `to_pyg`, `_graph_to_edge_tensors` and `_node_features_from_labels` all take their defaults from `DEFAULT_EDGE_FEATURES` / `DEFAULT_NODE_FEATURES` in `save/base.py`, no more None placeholders that get replaced in the body. The one None default left is `list_from_attribute`, which is a real optional (only branched on, never overwritten).

**Checklist**: type annotations on arguments and returns for every function in the package, and numpy-style docstrings throughout.

One more thing: main had moved under this branch (#117), so I merged main in and resolved the small `pyproject.toml` conflict (kept both `pyyaml` and `pyinstrument`). Full suite passes locally (230 tests). Ready for another look whenever you get a chance.

src/weather_model_graphs/save.py
leifdenby: hmm... but what controls `networkx`'s iteration order? I think it would be good to add a test here to ensure we have done this correctly, so we should add a test to demonstrate that sorting is necessary I think. What do you think?

src/weather_model_graphs/save.py
leifdenby: perfect, thanks!

src/weather_model_graphs/save.py
leifdenby: ah I see you have added a test already, https://github.com/prajwal-tech07/weather-model-graphs/blob/51b3e71506673e38a16dd7408c32ad275d582c11/tests/test_save.py#L412, fantastic! Ok, so reading what you wrote in the docstring for the edge ordering test I think we should just tweak the text above to detail that `networkx` uses insertion-ordering (and generates node indexes based on that order, right?) whereas we want the node indexes with the node labels we have provided on the `networkx.DiGraph` object. Is that right? Maybe we should add a further test to just check, without using any functionality in `weather-model-graphs` that if one creates a `nx.DiGraph` and adds edges not in order of the node labels, then the resulting edge-index tensor has the wrong node index values. What do you think? This has issue has been recurring many times, and I really think it would be good to document what we know about `network.DiGraphs` and conversion to torch edge-index tensors using `pyg`

src/weather_model_graphs/save/neural_lam/torch_tensors.py
leifdenby: every time I read your work I am blown away with how meticulously you work through your contributions. Fantatistic work as always @prajwal-tech07 - thank you!

tests/test_save.py
```diff
@@ -40,3 +48,439 @@ def test_save_to_pyg(list_from_attribute):
                 name=name,
                 list_from_attribute=list_from_attribute,
             )
+
+
+# ─── to_torch_tensors_on_disk: validated against neural-lam's own validator ───
```
leifdenby: these tests are great!

what do you think about splitting the new tests you are adding here that test the new wmg to neural-lam bridge into a separate test module? This file is getting quite long and I expect we will eventually remove the deprecated save functionality and so will delete the tests for that functionality too

leifdenby (commented):
Only two questions left :) Thanks again for this!

tests/test_save.py
prajwal-tech07: Good idea, will do. I'll move everything that tests the new bridge (the validator-backed spec tests, the `to_torch_tensors_on_disk` behaviour tests and the `_graph_to_edge_tensors` unit tests) into a separate module, and leave `test_save.py` with only the `to_pyg` tests plus the deprecation-warning test, so when the deprecated functionality is removed that whole file can go with it.

src/weather_model_graphs/save.py
prajwal-tech07: Yes, that's right, with one distinction worth spelling out. networkx iterates nodes and edges in insertion order, and pyg's `from_networkx` numbers nodes by that iteration order (its mapping is literally `zip(G.nodes(), range(N)))`, so the edge-index values end up meaning insertion positions, not our node labels. I verified this directly: inserting nodes as 5, 2, 9 and converting, the edge 5->2 comes out as [0, 1]. That's the recurring bug, and it's why the old `to_pyg` path needs `sort_nodes_in_graph` before conversion.

The new `_graph_to_edge_tensors` avoids that class of bug by construction: the index values come from the explicit sender/receiver maps built from sorted node labels, never from iteration order. So the sort in this function is about something else, the column order of the edges, so the same graph always serialises byte-identically. I'll reword the comment to make those two concerns explicit.

And agreed on the extra test, I'll add one that uses no wmg functionality at all: build an `nx.DiGraph` with edges added out of label order, convert with pyg, and show the resulting edge-index values don't match the labels (and that sorting nodes first fixes it). With a proper docstring it can serve as the reference for what we know about DiGraph -> torch edge-index conversion.

tests/test_save.py
prajwal-tech07: Done in 08d5ac2. All the bridge tests (the validator-backed spec tests, the `to_torch_tensors_on_disk` behaviour tests and the `_graph_to_edge_tensors` unit tests, plus the new networkx ordering pitfall test) now live in `tests/test_save_torch_tensors.py`. `tests/test_save.py` only keeps the `to_pyg` tests and the deprecation-warning test, with a module docstring noting it's expected to be deleted together with the deprecated functionality.

src/weather_model_graphs/save.py
prajwal-tech07: Done in 08d5ac2. Reworded the comment above the sort so the two concerns are spelled out separately: node index values (which this function gets from the sender/receiver maps built from sorted node labels, so iteration order can't affect them) and edge column order (which the sort fixes so the same graph always serialises byte-identically).

Also added the test you suggested, `test_pyg_from_networkx_indexes_by_insertion_order_not_labels`. It uses no wmg functionality at all: it builds an `nx.DiGraph` inserting nodes as 5, 2, 9 and shows that `from_networkx` encodes the edge 5->2 as (0, 1), i.e. insertion positions with no relation to the labels, and that inserting the nodes in sorted label order makes the indexes coincide with label rank. The docstring documents the whole pitfall plus the two ways wmg guards against it (sorting nodes before `from_networkx` in the deprecated path, explicit label-sorted maps in `_graph_to_edge_tensors`), so it can serve as the reference for this recurring issue.

leifdenby (changes requested):
One last thing: Can you add a changelog entry please? THen we can merge this :)

leifdenby (approved):
Perfect! All done 🥳

Issue #144, opened 2026-06-26
[Feat] Automated benchmark regression check in CI (PR vs main)


prajwal-tech07:
Spinning this out into its own issue (per the dev-meeting discussion) so the CI regression-check has a dedicated home. Builds on the benchmark scripts in #117 and the JSON output in #140 (cc @yuvraajnarula), and follows up the "Automated Performance Regression" part of #137.

Everything below runs entirely on GitHub-hosted runners - no external service or account required.

## Goal

On a PR, run the scaling benchmark on both the PR commit and current `main`, and post a comment showing whether the PR makes graph construction slower, faster, or the same - so regressions are visible before merge.

**Key design decision: compare on the same runner**

GitHub Actions runners are noisy - identical code can vary 20–50% between runs on different VMs. So comparing a PR against a baseline captured earlier on a different runner would produce frequent false alarms.

Instead I propose running `main` and the PR back-to-back in the same job, on the same runner, and comparing relative % change rather than absolute seconds. Same hardware for both sides → the noise largely cancels. This also matches the workflow described in the meeting.

**Correctness detail: hold the harness constant**

When we `git checkout main`, that reverts both the library and the benchmark script. To compare fairly we keep the benchmark script fixed (the PR's version) and only swap the library under test - i.e. run the same measurement against two installs of the code. Otherwise we'd be measuring with two different rulers.

## How it works

```yaml
on: pull_request (paths: src/.../create/**) + manual dispatch

1. checkout PR head → install lib → run scaling benchmark → pr.json
2. checkout main    → install lib → run same benchmark    → main.json
   (same runner, same job, same benchmark script)
3. compare.py main.json pr.json → markdown delta table
4. post / update a sticky PR comment with the result
```

Example comment:

| grid points | main | PR | Δ |
|-------------|------|----|----|
| 1,024 | 0.42s | 0.43s | +2% ✅ |
| 4,096 | 1.81s | 2.30s | +27% ⚠️ |

The only new code is a small `compare.py` (parse the two JSONs from #140, compute deltas, render the table, flag on threshold) plus the workflow YAML.

## Guardrails against noise

- Compare % change, never absolute time.
- Tolerance threshold — only flag a regression above e.g. +25% and consistent across grid sizes.
- Cap grid sizes to fit the CI time budget (small/medium grids, not 10⁷ nodes).
[...]

PR #140, opened 2026-04-21
feat: add memory profiling and JSON output to scaling benchmark

yuvraajnarula:
## Describe your changes

This PR extends the existing scaling benchmark (`graph_creation_scaling.py`) with memory profiling, machine‑readable output (JSON), and a separate memory scaling plot. It builds on the runtime scaling work in #117 and addresses the first part of #137 (memory profiling and baseline generation).

**Key additions:**

- `--track-memory` - enables peak memory tracking via `tracemalloc`.
- `--output-json <file>` -  saves raw results (grid points, runtime, peak memory) as JSON for later use (e.g., CI regression tests).
- `--output-plot-memory <file>` - generates a dedicated plot of peak memory vs number of grid points (requires `--track-memory`).
- Refactored the script into reusable functions:
  - `run_benchmark()` – returns a list of dicts with the measured data.
  - `plot_runtime_scaling()` - creates the runtime vs grid points plot.
  - `plot_memory_scaling()` - creates the memory vs grid points plot.

**Motivation and context:**

- Establishing baseline memory/runtime numbers on `main` is required for future performance regression testing.
- Machine‑readable output (JSON) enables programmatic consumption, which will be used in a follow‑up PR for CI regression tests.
- The memory scaling plot helps identify when graph generation becomes memory‑bound, critical for large global grids (e.g., 10⁷ nodes).

**Dependencies:**

- `tracemalloc` (standard library) - no new external dependencies.

## Issue Link

Closes #137 (first part - memory profiling and baseline tooling)

## Type of change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📖 Documentation

## Checklist before requesting a review

- [x] My branch is up-to-date with the target branch (`mllam/main`).
- [x] I have performed a self-review of my code.
- [x] For any new/modified functions/classes I have added docstrings.
- [x] I have placed in-line comments to clarify the intent.
- [ ] I have updated the documentation (README.md and docstrings).
- [ ] I have added tests that prove my fix is effective or that my feature works (existing tests pass; new functionality tested manually).
- [x] I have given the PR a name that clearly describes the change, written in imperative form: **"feat: add memory profiling and JSON output to scaling benchmark"**.
- [x] I have requested a reviewer and an assignee (assignee is responsible for merging).

## Checklist for reviewers

- [ ] the code is readable
- [ ] the code is well tested
- [ ] the code is documented (including return types and parameters)
- [ ] the code is easy to maintain

[...]

PR #147, opened 2026-07-17
Add automated benchmark regression check in CI (#144)

prajwal-tech07:
## Describe your changes

Adds an automated benchmark **regression check** in CI, implementing Phase 1 of #144.

On a pull request, the scaling benchmark is run on the **PR** and on its **base branch back-to-back on the same runner**, and a **sticky PR comment** reports the relative runtime change per grid size. Because both runs execute on the same VM and we compare **relative (%) change rather than absolute seconds**, the per-runner noise (which can be 20–50% across different GitHub-hosted runners) largely cancels.

**How the harness is held constant:** the benchmark script always comes from the PR checkout, and only the *library under test* is swapped between the two runs. The package uses a `src/` layout, so the working tree never shadows the installed library - installing the base-branch build over the PR build cleanly swaps the code being measured while the ruler (the benchmark harness) stays fixed.

Changes:
- **`tests/benchmarks/compare.py`** - standard-library-only script that parses two `--output-json` files (schema introduced in #140: `{grid_points, runtime_s, peak_memory_mb}`), computes per-grid-size % deltas, renders the Markdown comparison table, and flags any grid size above a configurable threshold. Informational by default (`--fail-on-regression` available for later).
- **`tests/benchmarks/test_compare.py`** - 17 unit tests covering the comparison logic (regression/improvement/threshold boundary, null memory, mismatched grid sizes, zero-baseline guard, malformed input, exit codes, output file).
- **`.github/workflows/benchmark-regression.yml`** - the same-runner A/B workflow. Uses `uv` for env setup, swaps only the library via `uv pip install --reinstall-package`, runs `compare.py`, and posts/updates the sticky comment via first-party `actions/github-script` (with a job-summary fallback so fork PRs still see results).

**Design notes / deliberate choices** (per the discussion in #144):
- The threshold **starts very low (0.1%)** on purpose - it's meant to be ratcheted up in follow-up runs once we've observed the runner's real noise floor, rather than guessing a large number upfront.
- The check is **non-blocking**, so a noisy perf signal doesn't train people to ignore CI. It can be promoted to a hard gate later once the threshold is proven.
- **Fork PRs** get a read-only token, so on forks the sticky comment is skipped and results go to the job summary instead. This is intentional - commenting on forks would require `pull_request_target`, which would run untrusted PR code with a write token.
- Memory-delta reporting and historical trend tracking are left for later phases.

**Dependencies:** no new runtime dependencies for the library. The workflow installs the package with its `visualisation` extra (matplotlib, already an optional dependency) to run the benchmark. `compare.py` is standard-library-only.

## Issue Link

Addresses #144 (Phase 1). Builds on the benchmark scripts from #117 and the JSON output from #140.

## Type of change

### DIFF
--- a/.github/workflows/benchmark-regression.yml
+++ b/.github/workflows/benchmark-regression.yml
@@ -1,0 +1,112 @@
+# Automated benchmark regression check (PR vs base branch).
+#
+# Design (see https://github.com/mllam/weather-model-graphs/issues/144):
+#   * Run the scaling benchmark twice, back-to-back, on the SAME runner: once
+#     against the PR's install of the library and once against the base branch's
+#     install. Only the library under test is swapped - the benchmark harness
+#     itself always comes from the PR checkout (the package uses a src/ layout,
+#     so the working tree never shadows the installed library).
+#   * Compare RELATIVE (%) change, not absolute seconds, so per-runner noise
+#     largely cancels.
+#   * Post the result as a single, updating ("sticky") PR comment. The check is
+#     informational and NON-BLOCKING for now; the threshold starts low and can
+#     be raised once the runner's real noise floor is known.
+
+name: benchmark regression
+
+on:
+  pull_request:
+    paths:
+      - "src/weather_model_graphs/create/**"
+      - "tests/benchmarks/**"
+      - ".github/workflows/benchmark-regression.yml"
+  workflow_dispatch:
+
+# Only the latest run per PR needs to finish; cancel superseded runs.
+concurrency:
+  group: benchmark-regression-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
+      # Kept small so the whole job stays within a few minutes. Tune here.
+      BENCH_ARGS: "--min-N 50 --max-N 200 --num-steps 4 --archetype keisler"
+      # Start low; raise once we've seen the runner's noise floor across a few
+      # real runs (per the discussion on #144).
+      THRESHOLD_PCT: "0.1"
+      BASE_REF: ${{ github.event.pull_request.base.ref || 'main' }}
+    steps:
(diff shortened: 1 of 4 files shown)


prajwal-tech07:
Self-reviewed and ready for review ✅

This implements **Phase 1** of #144 (automated benchmark regression check).

- `tests/benchmarks/compare.py` — parses two `--output-json` runs (schema from #140), computes per-grid-size % deltas, renders the sticky Markdown table, flags above a configurable threshold. Standard-library only.
- `tests/benchmarks/test_compare.py` — 17 unit tests (all green).
- `.github/workflows/benchmark-regression.yml` — same-runner A/B: benchmark the PR build, swap in the base-branch library (the `src/` layout keeps the PR's harness fixed), benchmark again, `compare.py`, post a sticky comment. **Informational / non-blocking.**

Notes:
- The `benchmark` check ran green on this PR itself, exercising the full A/B flow end-to-end.
- This is a **fork PR**, so the sticky-comment step falls back to the job summary (fork tokens are read-only) — by design; a same-repo PR posts the real comment.
- Threshold intentionally starts at **0.1%** so it can be calibrated up against the runner's real noise floor over the first few runs.

No rush given the timeline — happy for this to wait for review. cc @yuvraajnarula (builds on your JSON output).

.github/workflows/benchmark-regression.yml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Publish result to the job summary
+        if: always()
+        run: cat comment.md >> "$GITHUB_STEP_SUMMARY" || true
+
+      - name: Post or update sticky PR comment
+        # Only same-repo PRs get a writable token; fork PRs fall back to the job
+        # summary above. Never fail the job over the comment (informational).
+        if: github.event_name == 'pull_request'
+        continue-on-error: true
+        uses: actions/github-script@v7
```
leifdenby: I think we should use https://github.com/marocchino/sticky-pull-request-comment here instead

prajwal-tech07: Done - swapped to `marocchino/sticky-pull-request-comment@v3` in 90f918d

